PR #4, opened 2016-07-20
Add a rule to send email when scoring has value been reached

amidani-jahia:


services/src/main/java/org/apache/unomi/services/services/SegmentServiceImpl.java
```diff
@@ -669,6 +676,9 @@ private void updateExistingProfilesForScoring(Scoring scoring) {
                 scriptParams.put("scoringValue", element.getValue());
                 for (Profile p : persistenceService.query(element.getCondition(), null, Profile.class)) {
                     persistenceService.updateWithScript(p.getItemId(), null, Profile.class, script, scriptParams);
+                    Event profileUpdated = new Event("profileUpdated", null, p, null, null, p, new Date());
+                    profileUpdated.setPersistent(false);
+                    eventService.send(profileUpdated);
                 }
```
tdraier: The event will be sent multiple times per profile (actualy, for each matching scoring element) - could you optimize and send it only once ?

tdraier: It would also be nice to do the same in updateExistingProfilesForSegment()

PR #7, opened 2016-08-09
UNOMI-28 Upgrade to ElasticSearch 2.x

sergehuber:
Used this resource to perform changes:
https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_java_api_changes.html

A lot of changes are contained in this commit, notably :
- Moved data directory from KARAF_HOME/data to KARAF_HOME/elasticsearch/data
- Had to build a JSON parser to rewrite field names with "." to escape them with "**DOT**"
- Fixes on dates
- Replace all filter classes with query classes as these don't exist in ES 2.x any more
- Replace and/or ES clauses with bool must/should
- Had to replace deleteByQuery support to add plugin by default, since in ES 2.x they externalized the functionality
- Had to install Groovy and Expression modules since they are no longer part of the core

What is NOT done by this patch is the migration procedure


sergehuber:
I have removed the DOT parsing since in ES 2.4.x a startup flag allows to let dots be used in field names again
Also in this branch is the beginnings of the integration of the BulkProcessor API to speed up some bulk operations (such as profile updates)

sergehuber:
Close and replaced by PR #20

PR #26, opened 2017-02-27
UNOMI-79 : Implement save persona with sessions.

rizak:
Implemented the ability to save a personaWithSession object via UNOMI API.


sergehuber:
Thanks for providing all the comments on the methods.

PR #30, opened 2017-07-27
Weather update

dinesh1996:
Add new extension update Session with weather information based on user location


extensions/weather-update/core/src/main/java/org/apache/unomi/weatherupdate/actions/WeatherUpdateAction.java
```diff
@@ -0,0 +1,279 @@
...
+ * @author dsalhotra
+ */
+public class WeatherUpdateAction implements ActionExecutor {
+
+    private static Logger logger = LoggerFactory.getLogger(WeatherUpdateAction.class);
+
+    private static final double KELVIN = 273.15;
+    private static final double ROUND_TO_THE_TENTH = 0.5;
+    private static final double SECOND_TO_HOUR = 3.6;
+    private static final String MAIM_INFO_WEATHER = "main";
```
amidani-jahia: MAIN, Typo?

PR #40, opened 2017-12-08
Mapping properties

dinesh1996:


DameniMilo:
- Missing one placeholder for the CFG file
- rename mailChimpConnector.list.merge-fields to mailChimpConnector.list.merge-fields.mapping
- rename mailChimpConnector.list.merge-fields.activation to mailChimpConnector.list.merge-fields.activate
- rename listMergeField to listMergeFieldMapping
- rename mappingMergeFieldsActivation to isListMergeFieldsActivated
- Do not forget to rename associate getter/setter
- The function updateMergerPropertiesForList should be renamed to updateMCProfileProperties, the code to create the properties in the list should be extracted from this function to a private function named something like createPropertiesForList, add a javadoc for this specific function to explain why we need to do that.
- You should also create a private function named updateMCProfileProperties to be use by the addToMCList and therefore change the signature of the public function which should not take the actual last parameter.
- In the addToMCList ensure you add the profile or update the status before trying to update the profile with the new properties.
- getAllMergeFieldofOneMCList ??? should probably be renamed to getMCListProperties
- newMergeFields to mergeFields
- tabMergerField to mergeFieldInfo

PR #46, opened 2018-02-07
UNOMI-149 fix scoring condition strategy

Taybou:


Taybou:
If the threshold have a default value equal to 1, there is a small issue, that if we have a matching condition with the score value equal to 0

tdraier:
What is the point of having a condition with a score 0 ? And even if you had one, we would expect that had no influence on the result, and the content should not match. In your case it will match, which is rather inconsistent.

Taybou:
In this case, i prefer to keep the default value of the threshold to 0, and change the operator of comparison in the condition of IF statement to `>` instead of `>=`.

tdraier:
Mmm for me the threshold is the value at which the effects starts to happen, so _including_ the value itself, but it's more a matter of semantics :-)

Taybou:
I agree.

loomchild:
The new solution with default threshold = 1 looks good to me. I will run some manual tests.

loomchild:
After testing manually it seems to work correctly. @tdraier do you think we can merge this PR?

PR #47, opened 2018-02-07
UNOMI-150 Delete old ES index template during migration to 1.3.0

loomchild:
It causes problems since after migrating to a new version, there are two index templates in Elasticsearch matching the same index name pattern. Due to this, newly created monthly indices will use old configuration instead of new one.

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
@@ -32,7 +32,7 @@ import java.util.*;
 /**
  * @author dgaillard
  */
-public class MigrationTo130 implements Migration {
+public class MigrationTo121 implements Migration {
 
     private CloseableHttpClient httpClient;
     private CommandSession session;
@@ -47,7 +47,7 @@ public class MigrationTo130 implements Migration {
 
     @Override
     public Version getToVersion() {
-        return new Version("1.3.0");
+        return new Version("1.2.1");
     }
 
     @Override
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
@@ -52,17 +52,10 @@ public class MigrationTo121 implements Migration {
 
     @Override
     public void execute(CommandSession session, CloseableHttpClient httpClient, String esAddress) throws IOException {
-        try {
-            this.httpClient = httpClient;
-            this.session = session;
-            this.esAddress = esAddress;
-            migrateTags();
-        } catch (IOException e) {
-            if (httpClient != null) {
-                httpClient.close();
-            }
-            throw e;
-        }
+        this.httpClient = httpClient;
+        this.session = session;
+        this.esAddress = esAddress;
+        migrateTags();
     }
 
     private void migrateTags() throws IOException {
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
@@ -51,6 +51,11 @@ public class MigrationTo121 implements Migration {
     }
 
     @Override
+    public String getDescription() {
+        return "Migrate tags";
+    }
+
+    @Override
     public void execute(CommandSession session, CloseableHttpClient httpClient, String esAddress) throws IOException {
         this.httpClient = httpClient;
         this.session = session;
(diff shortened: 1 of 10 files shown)


loomchild:
I have chosen to add a second migration script to 1.3.0 version. This could cause problems to people who already executed a migration to 1.3.0, but in theory nobody should execute migration script of unreleased version in important (production-like) environments. If this is a problem, a simple solution would be to allow choosing which migration scripts to execute.

This is my first pull request, comments welcome.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo130IndexTemplate.java
```diff
@@ -0,0 +1,74 @@
...
+    public void execute(CommandSession session, CloseableHttpClient httpClient, String esAddress) throws IOException {
+        try {
+            this.httpClient = httpClient;
+            this.session = session;
+            this.esAddress = esAddress;
+            deleteOldIndexTemplate();
+        } catch (IOException e) {
+            if (httpClient != null) {
+                httpClient.close();
+            }
```
Taybou: I guess, it's better to move this close statement in the finally block

loomchild: I agree. I would say even more - managing the client shouldn't be migration script responsibility, it should be done in Migrate class that runs all the scripts.

loomchild: Done in https://github.com/apache/incubator-unomi/pull/47/commits/b463b11ef0d11203e72199fa569f013312cdc3a3, moved HTTP client resource management to main Migrate script.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo122.java
```diff
@@ -0,0 +1,67 @@
...
+        this.esAddress = esAddress;
+        deleteOldIndexTemplate();
+
+    }
+
+    private void deleteOldIndexTemplate() throws IOException {
+        String oldMonthlyIndexTemplate = "context_monthlyindex";
+        try {
+            ConsoleUtils.printMessage(session,"Deleting old monthly index template " + oldMonthlyIndexTemplate);
+            HttpUtils.executeDeleteRequest(httpClient, esAddress + "/_template/" + oldMonthlyIndexTemplate, null);
```
loomchild: No need to check if index template exists since DELETE operation will return HTTP code 404 in this case, and I am not treating it as an error.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
```diff
@@ -32,10 +32,11 @@
 /**
  * @author dgaillard
  */
-public class MigrationTo130 implements Migration {
+public class MigrationTo121 implements Migration {
```
sergehuber: There is no 1.2.1 released at Apache, I'm not sure this is a good idea to do this.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
```diff
@@ -46,35 +47,28 @@ public Version getFromVersion() {
 
     @Override
     public Version getToVersion() {
-        return new Version("1.3.0");
+        return new Version("1.2.1");
```
sergehuber: Same question here.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo122.java
```diff
@@ -0,0 +1,67 @@
...
+    private String esAddress;
+
+    @Override
+    public Version getFromVersion() {
+        return null;
+    }
+
+    @Override
+    public Version getToVersion() {
+        return new Version("1.2.2");
```
sergehuber: Again what is this 1.2.2 version ?
Maybe we could use date stamp or even commit stamps or build number instead ?

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo121.java
sergehuber: Maybe we should name these after what they do ? MigrateToNewTags / MigrateIndexTemplates ?

loomchild: I know it has not been released here, but on the other hand it has been released somewhere. It also does not break any existing functionality - all 1.2.x scripts will be executed when migrating to 1.3.0 after releasing it here.

This permits us to automate the process a bit, otherwise we will need to specify in the migration manual which scripts to run and which not.

PR #48, opened 2018-02-08
UNOMI-151 add send back bad request message 

dinesh1996:


wab/src/main/java/org/apache/unomi/web/ContextServlet.java
```diff
@@ -136,11 +136,17 @@ public void service(ServletRequest request, ServletResponse response) throws Ser
...
                 logger.error("Cannot read payload " + stringPayload, e);
                 return;
             }
             scope = contextRequest.getSource().getScope();
         }
+        if (httpMethod.equals("POST")){
+            ((HttpServletResponse)response).sendError(HttpServletResponse.SC_BAD_REQUEST, "Check logs for more details");
+            logger.error("Invalid Resource, please check the request");
+            return;
+        }
```
loomchild: So you want to reject only POST requests? Perhaps instead of defining which requests we don't support we should define which ones we support and throw the error otherwise? 

Also check the above treatment of OPTIONS http method above - isn't it a bit too late where we check it, shouldn't we group the checks together and use similar code?

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
```diff
@@ -136,11 +136,17 @@ public void service(ServletRequest request, ServletResponse response) throws Ser
             try {
                 contextRequest = mapper.readValue(factory.createParser(stringPayload), ContextRequest.class);
             } catch (Exception e) {
+                ((HttpServletResponse)response).sendError(HttpServletResponse.SC_BAD_REQUEST, "Check logs for more details");
```
loomchild: Shouldn't the HTTP code be more like 500 Internal Server Error? After all we don't know if the request is bad or we have another issue in our code / infra. Or perhaps it would be better to distinguish the user input errors (JsonParseException and JsonMappingException for example, didn't check in details) and respond with 400, otherwise respond with 500.

dinesh1996: The ticket is only aboute the post request, i asked to Serge.

loomchild: I think you meant to respond to my below comment and not this one, is that right? Could you also clarify what was his exact comment / explanation what are you trying to fix and put it in the JIRA ticket. Do we want to forbit all POST requests on this resource?

Another thing - as far as I know we create an UNOMI ticket with all the necessary details and reference it in commit message - could you do it please?

sergehuber: I don't understand what you are talking about @loomchild . The modifications in this commit only intend to make sure that an HTTP error code is sent when an error is detect. Before these changes the error code was not returned and the default code (200) was being sent back, which is exactly what we are trying to solve. But I do agree a ticket in the UNOMI JIRA should have been created first and used as a reference in this PR title. Please correct this @dinesh1996

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
sergehuber: see above comment. We are not changing the handling of the different requests, simply making sure errors that are logged as errors also return an HTTP error code.

loomchild: Yeah I understand the intention, but above code suggests otherwise - it sends an error when HTTP method is POST if I see correctly. I just wanted to confirm that is what we want. Perhaps this if condition should be in the catch block or something?

Also if that's what we want, I would move this if block a bit higher, close to the if block that checks for OPTIONS httpMethod (this is more question of taste).

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
loomchild: Here I am suggesting to perhaps respond with HTTP 400 Bad Request just when the error is caused by user request input data (catch just JsonParseException and JsonMappingException for example) and respond with HTTP 500 Internal Server Error in case of any other exception (NPE, IOException - since mapper.readValue() can throw it I suppose it communicates with something external).

I am not sure if this is a valid suggestion, since I don't know the code well yet. Perhaps we *want* to treat all errors in this line as HTTP 400 Bad request since they are always caused by used data (waiting for your confirmation):
`contextRequest = mapper.readValue(factory.createParser(stringPayload), ContextRequest.class);`

loomchild:
I see that you have changed the pull request title, but you'll also need to change the commit message and preferably keep both UNOMI-151 and the description.

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
loomchild: I am also saying that if we really want to block POST method, then we should probably also block DELETE, PATCH, PUT, etc., so it would be better to invert the condition and block anything else than GET and OPTIONS (whitelist instead of blacklist).

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
sergehuber: Ok that makes sense to me but it's not related to the HTTP method at all, rather the actual code executing in the context of an HTTP call. However this is extending the initial scope of this modification and I suggest we create a seperate issue for this so that we can access this already useful change.

wab/src/main/java/org/apache/unomi/web/ContextServlet.java
```diff
@@ -136,11 +137,17 @@ public void service(ServletRequest request, ServletResponse response) throws Ser
...
                 logger.error("Cannot read payload " + stringPayload, e);
                 return;
             }
             scope = contextRequest.getSource().getScope();
         }
+        if (httpMethod.equals("POST") && StringUtils.isBlank(stringPayload)){
+            ((HttpServletResponse)response).sendError(HttpServletResponse.SC_BAD_REQUEST, "Check logs for more details");
+            logger.error("Invalid Resource, please check the request");
+            return;
+        }
```
sergehuber: Could you explain the point of this code ? It was not, before modification, mandatory to have a payload and this modification makes it required. I don't think this is what is needed. All that we need is the modification on the error handling for the JSON parsing.

loomchild:
It works for me:
- by default Unomi responds 200 to MF
- when I forget sessionId parameter I get 400
- if I send invalid JSON in body, I get 400
- if I won't send body, I get 200, but I guess it's OK

Also thanks for chaning the commit message to indicate ticket number and rebasing.

All good to go for me!

sergehuber:
Ok thanks for the tests I'll merge this now.

PR #49, opened 2018-02-13
UNOMI-152 Fix monthly index New Year's bug

loomchild:
When formatting the date around new year's day, retrieving WEEK_YEAR can
yield next year or previous year as result. We should simply use YEAR.

More details: https://stackoverflow.com/questions/8686331/y-returns-2012-while-y-returns-2011-in-simpledateformat

### DIFF
--- a/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
+++ b/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
@@ -506,7 +506,7 @@ public class ElasticSearchPersistenceServiceImpl implements PersistenceService, 
     }
 
     private String getMonthlyIndexName(Date date) {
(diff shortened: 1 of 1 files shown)


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -506,7 +506,7 @@ public void bundleChanged(BundleEvent event) {
     }
 
     private String getMonthlyIndexName(Date date) {
-        String d = new SimpleDateFormat("-YYYY-MM").format(date);
+        String d = new SimpleDateFormat("-yyyy-MM").format(date);
```
loomchild: We execute this code quite ofter, and creating SimpleDateFormat could be expensive. However, we can't re-use it here I think, since it's not thread safe. 

I can optimize this code by changing it to this if you prefer:
`String d = "-" + (d.getYear() + 1900) + "-" + (d.getMonth() + 1)`

PR #50, opened 2018-02-20
UNOMI-165 Minor improvements in dev shell commands

loomchild:
- Consistent sorting of results by name
- Missing output to CSV
- Remove unnecessary --no-format parameter
- Various style fixes

### DIFF
--- a/tools/shell-dev-commands/src/main/java/org/apache/unomi/shell/commands/ActionListCommand.java
+++ b/tools/shell-dev-commands/src/main/java/org/apache/unomi/shell/commands/ActionListCommand.java
@@ -32,17 +32,14 @@ import java.util.Collection;
 @Command(scope = "action", name = "list", description = "This will list all the actions deployed in the Apache Unomi Context Server")
 public class ActionListCommand extends OsgiCommandSupport{
 
-    DefinitionsService definitionsService;
+    private DefinitionsService definitionsService;
+
+    @Option(name = "--csv", description = "Output table in CSV format", required = false, multiValued = false)
+    boolean csv;
 
     public void setDefinitionsService(DefinitionsService definitionsService) {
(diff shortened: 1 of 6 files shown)


tools/shell-dev-commands/src/main/java/org/apache/unomi/shell/commands/ActionListCommand.java
```diff
@@ -57,14 +54,14 @@ protected Object doExecute() throws Exception {
         DataTable dataTable = new DataTable();
 
         for (ActionType actionType : allActions) {
-            ArrayList<Object> rowData = new ArrayList<Object>();
+            ArrayList<Comparable> rowData = new ArrayList<>();
```
loomchild: I left it as Comparable, since we later cast to Comparable, so this is safer IMHO.

tools/shell-dev-commands/src/main/java/org/apache/unomi/shell/commands/ActionListCommand.java
```diff
@@ -57,14 +54,14 @@ protected Object doExecute() throws Exception {
...
             rowData.add(actionType.getItemId());
             rowData.add(actionType.getMetadata().getName());
             rowData.add(StringUtils.join(actionType.getMetadata().getSystemTags(), ","));
             dataTable.addRow(rowData.toArray(new Comparable[rowData.size()]));
         }
 
-        dataTable.sort(new DataTable.SortCriteria(2, DataTable.SortOrder.ASCENDING));
+        dataTable.sort(new DataTable.SortCriteria(1, DataTable.SortOrder.ASCENDING));
```
loomchild: Here the ordering was wrong, since columns are 0-indexed.

PR #51, opened 2018-02-26
UNOMI-168 Fix property type IT

loomchild:
Now we retrieve property types from DB every 5 seconds
(see 681384714130e49d747c163c5c3ac70774263e59), so we need longer waits
when running integration tests.

### DIFF
--- a/services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
+++ b/services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
@@ -39,6 +39,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.URL;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ProfileServiceImpl implements ProfileService, SynchronousBundleListener {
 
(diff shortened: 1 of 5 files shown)


sergehuber:
As discussed over the phone, I think it would be better to fix the problem on the updates of the cached structures inside the ProfileServiceImpl.

services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
```diff
@@ -883,6 +937,7 @@ private void loadPredefinedPropertyTypes(BundleContext bundleContext) {
                 logger.error("Error while loading properties " + predefinedPropertyTypeURL, e);
             }
         }
+        loadPropertyTypesFromPersistence();
```
loomchild: I am not sure about this one, because the new values might still not be available in ES anyway. Is it causing any preactical trouble to initially have empty cache?

services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
```diff
@@ -862,6 +923,7 @@ private void loadPredefinedPropertyTypes(BundleContext bundleContext) {
             return;
         }
 
+        List<PropertyType> allPropertyTypes = new ArrayList<>();
```
sergehuber: This variable name is confusing I think it would be better to call it bundlePropertyTypes

services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
```diff
@@ -39,9 +39,99 @@
 import java.io.IOException;
 import java.net.URL;
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ProfileServiceImpl implements ProfileService, SynchronousBundleListener {
 
+    private static class PropertyTypes {
```
sergehuber: Maybe we could add a comment on this class that explains why we are using copies on write: basically to make sure that we don't impact performance with contention and that we insure a consistent state as much as possible.

loomchild: OK, I will do that.

services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
loomchild: You are right, I will do that.

PR #53, opened 2018-03-01
UNOMI-169 Run integration tests before deployment

loomchild:
To achieve this I introduced a dependency of package module on itests
module. It does not build if integration-tests profile is not activated.

### DIFF
--- a/package/pom.xml
+++ b/package/pom.xml
@@ -99,6 +99,14 @@
             <type>xml</type>
             <scope>runtime</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.unomi</groupId>
+            <artifactId>unomi-itests</artifactId>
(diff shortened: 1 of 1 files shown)


sergehuber:
So this means that we cannot build the package without running the integration tests ? That seems like quite a limitation during development where we might be rebuilding the package to test some changes, but we don't necessarily want to run the integration tests at this point. 

Also I wonder if this isn't just putting a requirement that the itests must have run once previously, but not necessarily in this current build, and therefore old itests might have been ok but not the current ones ?

loomchild:
No, we can build a package without integration tests. If you don't specify -Pintegration-test profile, the tests won't run (checked). I also experimented with `optional` dependency, but in the end it wasn't necessary.

As for old tests, I don't know. When I run `mvn clean install` it runs them every time, I will try without clean.

PR #54, opened 2018-03-05
fix unknown protocol: wrap

oliverlietz:


sergehuber:
Wow thanks Oliver for this I wasn't expecting it ! I will test and merge asap !

sergehuber:
Ok I've merged the PR because it improves things but I still have a problem with the resolver that is looping.

PR #61, opened 2018-05-03
UNOMI-179 Event Persistence Non-transient

donald-kortx:
Changed the persistence field to be non-transient in the Event class.
Used ESEventMixIn to prevent that value from being saved in ES.  Updated
the Event constructor to accept persistent.

This feature allows the user to pass "persistent": false with their
events if the intent is only to trigger a rule / action downstream.

If "persistent" is not set it defaults to true.

### DIFF
--- a/api/src/main/java/org/apache/unomi/api/Event.java
+++ b/api/src/main/java/org/apache/unomi/api/Event.java
@@ -113,6 +113,7 @@ public class Event extends Item implements TimestampedItem {
      * @param target     the target of the event if any
      * @param timestamp  the timestamp associated with the event if provided
      * @param properties the properties for this event if any
+     * @param persistent specifies if the event needs to be persisted
      */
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/apache/unomi/api/Event.java
```diff
@@ -114,8 +114,9 @@ public Event(String eventType, Session session, Profile profile, String scope, I
      * @param timestamp  the timestamp associated with the event if provided
      * @param properties the properties for this event if any
      */
-    public Event(String eventType, Session session, Profile profile, String scope, Item source, Item target, Map<String, Object> properties, Date timestamp) {
+    public Event(String eventType, Session session, Profile profile, String scope, Item source, Item target, Map<String, Object> properties, Date timestamp, boolean persistent) {
```
sergehuber: Could you please add a @param description for the parameter ?

sergehuber:
Thanks for the changes, I've merged this pull request !

PR #65, opened 2018-08-03
UNOMI-195 fix issue of retrieve totalhits when aggregations equals null

Taybou:


DameniMilo:
We won't merge this pull request as it changes the way aggregate works, there is no aggregation result so we do not return filtered data. instead you can use the /count end point to achieve what you want to do.

sergehuber:
Taybou can you please close this PR because it seems I can't do it from this side ?

Taybou:
Yes of course.

PR #73, opened 2018-12-19
.identify() does not send events to dmp

shpetimshala:
Events are not sent to unomi when I tried to identify users, code was commented

Snippet that trigger this event:

```
unomiTracker.identify("userId", {
    email: "user@email.com"
});
```

### DIFF
--- a/extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
+++ b/extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
@@ -170,9 +170,7 @@ Unomi.prototype.processReferrer = function() {
  * @param {Identify} identify
  */
 Unomi.prototype.onidentify = function(identify) {
-    console.log('onidentify');
-    console.log(identify);
-    // this.collectEvent(identify.json());
+    this.collectEvent(identify.json());
 };
 
 /**
--- a/manual/src/main/asciidoc/clustering.adoc
+++ b/manual/src/main/asciidoc/clustering.adoc
@@ -14,70 +14,26 @@
 === Cluster setup
 
 Apache Karaf relies on Apache Karaf Cellar, which in turn uses Hazelcast to discover and configure its cluster.
-You just need to install multiple context servers on the same network, and then (optionally) change the Hazelcast
- configuration in the following file :
+
+You can control most of the important clustering settings through the centralized configuration file at
+
+    etc/unomi.custom.system.properties
+
+And notably using the following properties:
+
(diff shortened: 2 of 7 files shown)


sergehuber:
Sorry I didn't see this PR sooner. I'm concerned about the security here. Normally an identify should be a secure operation because it can merge with other profile. 

How are you using this identify ? What type of event are you sending ?

lhwong:
https://segment.com/docs/spec/identify/
"Our recommendation for when and how often you should call identify is as follows:

After a user registers
After a user logs in
When a user updates their info (eg changes or adds a new address)"


In this case, should I use
1. "(Preferred) Use you own custom event(s) to send data you want to be inserted in a profile, and use rules to map the event data to the profile." as given under section 3.2.3. HOW TO UPDATE A PROFILE FROM THE PUBLIC INTERNET (https://unomi.apache.org/manual/latest/index.html#_how_to_update_a_profile_from_the_public_internet)
or
2. "performs an server-to-server call to send the login event to Apache Unomi using the Unomi key." as given under section 3.2.8. HOW TO SEND A LOGIN EVENT TO UNOMI (https://unomi.apache.org/manual/latest/index.html#_how_to_send_a_login_event_to_unomi)

To comply with the integration implementation of analytics.js for Javascript client, I believe the option 1 (invoke call from Javascript) seems to be the way. 


Then, "Setup a rule that can check a profile property to see if profiles can be merged on an universal identifier such as an email address by calling analytics.identify([userId], [traits], [options], [callback]). The example of rule is similar to https://github.com/apache/unomi/blob/master/samples/login-integration/src/main/resources/META-INF/cxs/rules/exampleLogin.json
This rule calls an action called :
mergeProfilesOnPropertyAction
What this means is that upon successful login using an email, Unomi will look for other profiles that have the same email and merge them into a single profile." 


However, this doesn't seems to solve the security concern you have: "Normally an identify should be a secure operation because it can merge with other profile". How can we call identity from public internet and merge profile based on userId?

sergehuber:
That's a very good question actually. I guess in this case the best thing is to be careful what kind of data you are storing in the profile to make sure that even if a merge happens no critical data can be retrieved. 

I've been working on a way to "protect" properties that would only be accessible if a user is authentified but this is not yet available in Unomi.

sergehuber:
Sorry I didn't get back on this. After reflection I think it could be ok to send an identify event if a merge is not necessarily trigger. This would just augment the profile. 

But this cannot be merged as is because we need to build a proper event for Unomi to process.

sergehuber:
I've implemented proper support for this in this PR:
https://github.com/apache/unomi/pull/115

I'm therefore closing this PR. Thanks for your contributions and please don't hesitate to continue the discussion on the other PR.

PR #78, opened 2019-03-12
UNOMI-220 eliminates leading and trailing spaces into ip value

Taybou:
Escape spaces if exists in the value of thirdparty.provider1.ipAddresses


sergehuber:
Thanks for the PR, it has now been merged.

PR #80, opened 2019-04-01
[UNOMI-224] fix an API bug which may crash the program.

bd2019us:
Fix: [#UNOMI-224](https://issues.apache.org/jira/browse/UNOMI-224) to avoid crash of program.


sergehuber:
Thanks for the PR, it has now been merged.

PR #81, opened 2019-04-01
Clean incubator urls

fpapon:
@sergehuber this PR clean some incubator URLs


sergehuber:
Thanks for catching these ! I will merge now

PR #92, opened 2019-07-12
Fix for Jira UNOMI-238

micheleriva:
Hi,
I've added a small fix for the Jira UNOMI-238.
I declared a new variable on line 1001 because I wasn't able to use `StringUtils. equalsAny` which was introduced with `StringUtils` version 3.5 ([source](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#equalsAny-java.lang.CharSequence-java.lang.CharSequence...-)).

Best regards
Michele

### DIFF
--- a/services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
+++ b/services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
@@ -997,6 +997,9 @@ public class ProfileServiceImpl implements ProfileService, SynchronousBundleList
     private boolean merge(Map<String, Object> target, Map<String, Object> object) {
         boolean changed = false;
         for (Map.Entry<String, Object> newEntry : object.entrySet()) {
+
+            String packageName = newEntry.getValue().getClass().getPackage().getName();
+
             if (newEntry.getValue() != null) {
                 if (newEntry.getValue() instanceof Collection) {
                     target.put(newEntry.getKey(), newEntry.getValue());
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java
```diff
@@ -997,6 +997,9 @@ public void bundleChanged(BundleEvent event) {
     private boolean merge(Map<String, Object> target, Map<String, Object> object) {
         boolean changed = false;
         for (Map.Entry<String, Object> newEntry : object.entrySet()) {
+
+            String packageName = newEntry.getValue().getClass().getPackage().getName();
```
fpapon: @micheleriva if `newEntry.getValue()` is null, it will throw a NPE.

micheleriva: Hi @fpapon 	,
I actually used the original Unomi code (see [here](https://github.com/apache/unomi/blob/master/services/src/main/java/org/apache/unomi/services/services/ProfileServiceImpl.java#L1012)), in order to retrive that value, I didn't want to upset the codebase

fpapon: @micheleriva yes but it's better to put this line after the test `if (newEntry.getValue() != null) {`
=> https://github.com/apache/unomi/pull/92/files#diff-b144601886f209ff7ba78b31f594f199R1003

fpapon: NPE is tested after so in the actual code it works.
=>https://github.com/apache/unomi/pull/92/files#diff-b144601886f209ff7ba78b31f594f199R1003

micheleriva: Ok I get your point, my bad. I'm sending an updated commit, thanks for your review

fpapon: Great ;) Thanks for your PR!

PR #100, opened 2019-09-25
fix typo in recipes

pragnesh:
I have found some issue with documentation. This pull request fix some of the issue.


sergehuber:
Thanks for catching these and correcting them !

PR #109, opened 2019-11-13
[UNOMI-256] Add device brand and device name to user agent detector

kressaty:
The useragent service does a great job of grabbing more information than is currently persisted, and this adds the device name (`iPhone 7`) and device brand (`Apple`) to the session information - useful for measurement and specific targeting needs.


fpapon:
@kressaty Thanks for this PR!
I created a JIRA just for tracing the release note:
https://issues.apache.org/jira/browse/UNOMI-256

PR #115, opened 2019-11-28
UNOMI-258 Add support for identify() and track() methods in web-tracker

sergehuber:
This commits add basic support to make sure that the identify and track methods do something and send back events to Unomi.
This commit does not contain any documentation but it does add an example to the web tracker example page.
You can view the generated events using the event-tail and event-view commands on the server.

Signed-off-by: Serge Huber <shuber@apache.org>


sergehuber:
Thanks for the review Kevan !

PR #118, opened 2019-12-03
UNOMI-225 ElasticSearch 7 fix for changes in the total hits

sergehuber:
- Replaced total hits with count API requests when possible
- For aggregations replaced with a manual sum of all buckets
- Added in the PartialList the relation for the total hits so that users of the API can know if the total size is EQUAL or GREATER_THAN_OR_EQUAL_TO

Signed-off-by: Serge Huber <shuber@apache.org>

### DIFF
--- a/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
+++ b/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
@@ -26,6 +26,7 @@ import org.apache.http.client.CredentialsProvider;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.apache.lucene.search.TotalHits;
 import org.apache.unomi.api.Item;
 import org.apache.unomi.api.PartialList;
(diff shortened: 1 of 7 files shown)


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -1346,13 +1349,22 @@ private long queryCount(final QueryBuilder filter, final String itemType) {
...
+                return response.getCount();
+                /*
                 SearchRequest searchRequest = new SearchRequest(getIndexNameForQuery(itemType));
                 SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
                 searchSourceBuilder.query(filter);
                 searchSourceBuilder.size(0);
                 searchRequest.source(searchSourceBuilder);
                 SearchResponse response = client.search(searchRequest, RequestOptions.DEFAULT);
                 return response.getHits().getTotalHits().value;
+                 */
```
Taybou: Maybe, we should drop this  code in the block comment.

sergehuber: Indeed you are right

sergehuber: I've removed it

PR #119, opened 2019-12-04
UNOMI-225: implement ssl communication support with ES7

jkevan:


sergehuber:
Thanks for the contribution looks good to me

PR #125, opened 2020-02-07
fix(elasticsearch-7): remove type from ids query, type is deprecated

lyogev:
Index type is deprecated, also now all types are _doc, this causes past event condition to not function.


jbonofre:
It's part of ongoing change, I think it's a duplicate.

jbonofre:
Yeah, my bad, the duplicate was with one of my local branche. +1 to merge.

PR #128, opened 2020-02-12
feat(pom.xml): bump karaf version to 4.2.8

RonBarabash:
Hey, 
Karaf references Maven Central via HTTP which stopped working today as per https://support.sonatype.com/hc/en-us/articles/360041287334-Central-501-HTTPS-Required
This is fixed in version 4.2.8
issue is attached here:
https://issues.apache.org/jira/browse/KARAF-6599


sergehuber:
Thanks for the contribution, looks good to me.

PR #132, opened 2020-02-28
UNOMI-281 Add lastUpdated system property to Profiles

sergehuber:
- This patch adds the lastUpdated system property to profiles
- Added as a system property to avoid any issues with migration. If it doesn't exist in existing installations it will simply be added the next time a profile is updated.
- Any profile modification (including changing consents, adding new segments or scoring plans) will update the last update date on profiles
- This patch also includes some various minor improvements to logging, that were used while developing this feature

### DIFF
--- a/api/src/main/java/org/apache/unomi/api/Profile.java
+++ b/api/src/main/java/org/apache/unomi/api/Profile.java
@@ -132,6 +132,20 @@ public class Profile extends Item {
     }
 
     /**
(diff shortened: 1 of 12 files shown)


Taybou:
Great work @sergehuber!

sergehuber:
Thanks for the (very) quick reviews guys ! I'll merge now.

PR #134, opened 2020-03-10
UNOMI-284 Make mapping total fields limit configurable

sergehuber:
For some usages the default limit (1000) is too low. This modification makes it configurable, especially for dynamically created indices.


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -139,8 +139,10 @@
     private String indexPrefix;
     private String monthlyIndexNumberOfShards;
     private String monthlyIndexNumberOfReplicas;
+    private String monthlyIndexMappingTotalFieldsLimit;
     private String numberOfShards;
     private String numberOfReplicas;
+    private String indexMappingTotalFieldsLimit;
```
Taybou: Why you didn't add this property to the internal create index mapping?

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -1056,7 +1056,8 @@ private void internalCreateIndex(String indexName, String mappingSource) throws
         createIndexRequest.settings("{\n" +
                         "    \"index\" : {\n" +
                         "        \"number_of_shards\" : " + numberOfShards + ",\n" +
-                        "        \"number_of_replicas\" : " + numberOfReplicas + "\n" +
+                        "        \"number_of_replicas\" : " + numberOfReplicas + ",\n" +
+                        "        \"index.mapping.total_fields.limit\" : " + monthlyIndexMappingTotalFieldsLimit + "\n" +
```
Taybou:
```suggestion
                        "        \"index.mapping.total_fields.limit\" : " + indexMappingTotalFieldsLimit + "\n" +
```

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
sergehuber: It was a mistake. I added it (with a bad copy-paste) but I committed your fix. Thanks for catching this (twice !)

PR #136, opened 2020-03-10
UNOMI-286 Event type definitions & set value type for property types

sergehuber:
This commit adds the following:
- Event type definitions : possibility to define event types and retrieve them through the REST API.
- Set value type for property types. Make it possible to define complex property types that may include other property types and therefore define complex object structures to be accepted as property types. Event types may also use this possibility to define their complete structure.

This PR is not yet complete but opened for prilimnary review & discussion.


sergehuber:
Closing this as it was implement in another PR#168 (part of UNOMI-352)

PR #147, opened 2020-04-17
UNOMI-300 Fix npm dependencies for web tracker

sergehuber:
- Upgrade to analytics.js 3.10.1
- Integrate Javascript compilation with Maven using frontend-maven-plugin
- Fix integration to work with version 3.10.1
- Cleanup of generated files


extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
```diff
@@ -19,23 +19,26 @@
 var integration = require('@segment/analytics.js-integration');
 var extend  = require('extend');
 
-var Unomi = module.exports = integration('Apache Unomi')
+var Unomi = (module.exports = integration('Apache Unomi')
```
Taybou, line 22:
```suggestion
var Unomi = module.exports = integration('Apache Unomi')
```

extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
```diff
@@ -19,23 +19,26 @@
...
+    .global('cxs')
     .assumesPageview()
     .readyOnLoad()
-    .global('cxs')
     .option('scope', 'systemscope')
     .option('url', 'http://localhost:8181')
     .option('timeoutInMilliseconds', 1500)
     .option('sessionCookieName', 'unomiSessionId')
-    .option('sessionId');
+    .option('sessionId'));
```
Taybou, line 30:
```suggestion
    .option('sessionId');
```

extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
```diff
@@ -19,23 +19,26 @@
...
 /**
  * Initialize.
  *
  * @api public
  */
-Unomi.prototype.initialize = function(page) {
+Unomi.prototype.initialize = function() {
     var self = this;
+
+    console.info('[UNOMI] Initializing...', arguments);
```
Taybou, line 40: where is the **`arguments`** variable declared?

sergehuber: This is a standard built-in variables exposed by functions, see : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
sergehuber: This was done this way because all the other analytics.js integrations use this format.

Taybou: I don't see any reason to add parentheses here.

We assume that js var is passed as a reference, and 
```
var x = y = "toto"
// x will be "toto"
// y will be "toto"
```

extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
Taybou: Yes, but the initialize function doesn't have any parameters at the moment

sergehuber: Javascript allows passing arguments to functions that don't declare them. This log is here to check what the library is sending, as it has changed over time and is expected to change again.

extensions/web-tracker/javascript/src/analytics.js-integration-apache-unomi.js
sergehuber: Not sure why this was done, but probably for legibility. Again here we are re-using the patterns in the analytics.js source code.

sergehuber: Here are examples:
https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/adobe-analytics/lib/index.js
https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/facebook-pixel/lib/index.js
https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/adobe-target/lib/index.js

PR #148, opened 2020-04-24
UNOMI-327 Various quality issues

sergehuber:
- Change heap size from 512MB (default) to 2GB
- Catch ElasticsearchStatusException and in the case of an index not found we simply return null, in all other cases we raise the exception
- Fix scroll identifier leak when query returns no results.

### DIFF
--- a/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
+++ b/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
@@ -38,6 +38,7 @@ import org.apache.unomi.metrics.MetricsService;
 import org.apache.unomi.persistence.elasticsearch.conditions.*;
 import org.apache.unomi.persistence.spi.PersistenceService;
 import org.apache.unomi.persistence.spi.aggregate.*;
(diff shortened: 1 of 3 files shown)


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -688,6 +690,12 @@ public T execute(Object... args) throws Exception {
                             return null;
                         }
                     }
+                } catch (ElasticsearchStatusException ese) {
+                    if (ese.status().equals(RestStatus.NOT_FOUND)) {
+                        // this can happen if we are just testing the existence of the item, it is not always an error.
+                        return null;
+                    }
+                    throw new Exception("Error loading itemType=" + clazz.getName() + " itemId=" + itemId, ese);
```
Taybou, line 698:
```suggestion
                    throw new Exception("Error loading itemType=" + class.getName() + " itemId=" + itemId, ese);
```

sergehuber: I'm sorry but this suggestion is not correct because clazz is an argument to the method, not a typo :)

PR #150, opened 2020-04-28
UNOMI-330 Events non-persistant by default

sergehuber:
Fixed by making them persistent by default.


api/src/main/java/org/apache/unomi/api/Event.java
```diff
@@ -100,7 +100,7 @@ public Event(String itemId, String eventType, Session session, Profile profile,
      * @param timestamp the timestamp associated with the event if provided
      */
     public Event(String eventType, Session session, Profile profile, String scope, Item source, Item target, Date timestamp) {
-        this(eventType, session, profile, scope, source, target, null, timestamp, false);
+        this(eventType, session, profile, scope, source, target, null, timestamp, true);
```
Taybou, line 103: It's quite confusing but true for persistent and the comment says non-persistent!

PR #155, opened 2020-05-03
UNOMI-334 Docker Maven integration

sergehuber:
This PR contains a Maven - Docker integration which brings the following benefits:
- Version is automatically managed by Maven
- Docker-compose is integrated with Maven, making it very easy to build and start a full Unomi environment by just using: `mvn clean install` and then `mvn docker:start` in the Docker sub-project
- Docker push is also possible directly using `mvn docker:push` (it will use the project version as a tag, which should make it also easy to handle releases).

### DIFF
--- a/docker/pom.xml
+++ b/docker/pom.xml
@@ -73,8 +73,7 @@
                     <images>
                         <image>
                             <alias>unomi</alias>
-                            <name>apache/unomi:latest</name>
-
+                            <name>apache/unomi:${project.version}</name>
                             <external>
                                 <type>compose</type>
                                 <basedir>${basedir}</basedir>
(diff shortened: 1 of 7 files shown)


docker/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+                                    <type>tar.gz</type>
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>docker-maven-plugin</artifactId>
```
jbonofre, line 70: FYI, in my branch, I replaced this with karaf-maven-plugin (as unomi is now a "real" karaf custom distribution).

sergehuber: You replaced the docker-maven-plugin with the karaf-maven-plugin ? Or the dependency on the distribution?

PR #158, opened 2020-05-11
Make it possible to white list and black list classes used in OGNL and MVEL2 scripting languages

sergehuber:


fpapon:
@sergehuber may be we should mention this improvement in the manual?

sergehuber:
Yes I was wondering about this. Not sure where to put it though.

sergehuber:
Ok I've added some documentation in the manual in the configuration section.

PR #160, opened 2020-05-17
Unomi-336 fix past event count using custom timestamp

nlevitsky:
JIRA task -
https://issues.apache.org/jira/projects/UNOMI/issues/UNOMI-336

Check the event's timestamp before increasing the pastEvent counter by 1


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/SetEventOccurenceCountAction.java
```diff
@@ -80,7 +86,16 @@ public int execute(Action action, Event event) {
             pastEvents = new LinkedHashMap<>();
             event.getProfile().getSystemProperties().put("pastEvents", pastEvents);
         }
-        pastEvents.put((String) pastEventCondition.getParameter("generatedPropertyKey"), count + 1);
+
+        //increase the counter by 1 if the current event fulfils the time condition
```
sergehuber, line 90: I think this comment needs to be a bit clearer, something like:
```suggestion
        // Only increase the counter by 1 if the current event is in the now-numberofDays to now range
```

nlevitsky: Changed

nlevitsky: Added integration tests

PR #161, opened 2020-05-22
UNOMI-338 avoid update date with invalid custom timestamp

Taybou:


sergehuber:
Thanks for the PR ! Looks good to me.

PR #163, opened 2020-06-04
UNOMI-343 Fix integrations tests that are currently broken

pmi:


itests/src/test/java/org/apache/unomi/itests/graphql/GraphQLWebSocketIT.java
```diff
@@ -105,7 +111,11 @@ public void onWebSocketConnect(Session sess) {
 
         @Override
         public void onWebSocketText(String message) {
-            this.emitter.onNext(message);
```
pmi, line 108: NPE was here, because callback on line 97 was not invoked yet

sergehuber:
Thank you for the fixes !

PR #164, opened 2020-06-05
UNOMI-341 - Refactor ServiceManager

sergehuber:
In this PR the idea is to avoid having a ServiceManager class that is hardcoded to specific services, so that it becomes easy to access new services, especially if plugins are added overtime.


pmi:
Looks good !

PR #167, opened 2020-06-14
fix(ContextServletIT) - fix failing integration test

nlevitsky:
There was an error when checking if the event's timestamp is in the future. 
When calling Duration.toDays(), negative durations are rounded to 0. This caused future events to be calculated as if they match the "pastEvent" condition


sergehuber:
Wow nice catch I didn't see that. Thanks for the contribution.

PR #168, opened 2020-06-24
UNOMI-352 Implement event type registry for GraphQL

pmi:


api/src/main/java/org/apache/unomi/api/services/EventTypeRegistry.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.unomi.api.services;
+
+import org.apache.unomi.api.EventType;
+
+import java.util.Collection;
+
+public interface EventTypeRegistry {
```
sergehuber, line 24: Please add basic documentation on this class

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
```diff
@@ -59,7 +59,17 @@
 
     private BundleContext bundleContext;
     public DefinitionsServiceImpl() {
-
+        // let's add the built-in value types
```
sergehuber, line 62: I wonder if we should add all the built-in value types programmatically like this or move them all to JSON files. But having a mixture of this doesn't seem coherent.

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
```diff
@@ -160,7 +170,7 @@ private void processBundleStop(BundleContext bundleContext) {
         if (bundleContext == null) {
             return;
         }
-        List<PluginType> types = pluginTypes.get(bundleContext.getBundle().getBundleId());
+        List<PluginType> types = pluginTypes.remove(bundleContext.getBundle().getBundleId());
```
sergehuber, line 173: Is this a bug fix ?

services/src/main/java/org/apache/unomi/services/impl/events/EventServiceImpl.java
```diff
@@ -211,10 +226,61 @@ private void getEventProperties(Map<String, Map<String, Object>> mappings, List<
...
+        }
+        return properties;
+    }
+
+    private ValueType mappingTypeToValueType(String mappingType) {
+        if ("text".equals(mappingType)) {
+            return definitionsService.getValueType("string");
+        } else if ("date".equals(mappingType)) {
+            return definitionsService.getValueType("date");
+        } else if ("long".equals(mappingType)) {
```
sergehuber, line 266: I'm not sure whether we should translate long to integer or introduce a "long" in Unomi. I've been playing with the second idea.

services/src/main/resources/META-INF/cxs/events/common.json
```diff
@@ -0,0 +1,34 @@
+{
+  "propertyTypes": [
+    {
+      "itemId": "id",
+      "type": "id"
+    },
+    {
+      "itemId": "cdp_source"
```
sergehuber, line 8: Should we really have the cdp fields here ? I thought we would have the Event.class fields here and then create a layer to translate them to cdp in the GraphQL code which would seem more coherent.

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
pmi: That was done by you in UNOMI-286 that I based my work on, so I can only guess :)

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
pmi: That came from UNOMI-286 that I based my work on, but I will remove that, as we have set property type defined in json. Don't think we need _unknown_ though.

services/src/main/java/org/apache/unomi/services/impl/events/EventServiceImpl.java
pmi: Yeah, but maybe it's better to do it as a separate task ?

services/src/main/resources/META-INF/cxs/events/common.json
pmi: Removed common.json completely

api/src/main/java/org/apache/unomi/api/services/EventTypeRegistry.java
pmi: Done

api/src/main/java/org/apache/unomi/api/services/EventService.java
```diff
@@ -82,9 +79,19 @@
      * Retrieves the list of available event properties.
      *
      * @return a list of available event properties
+     * Retrieves an event type
```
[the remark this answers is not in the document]
pmi: Fixed

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
sergehuber: Ok yes unknown is not needed let's get rid of it.

services/src/main/java/org/apache/unomi/services/impl/definitions/DefinitionsServiceImpl.java
sergehuber: ok sorry didn't realize that :)

services/src/main/java/org/apache/unomi/services/impl/events/EventServiceImpl.java
sergehuber: agreed but we will need to remember it somehow

sergehuber:
Thanks for the PR, I just merged it !

PR #169, opened 2020-06-25
Add an environment variable

nlevitsky:
Add an environment variable for Elasticsearch refresh after saving a profile


nlevitsky:
> Am I missing something or is this PR not complete ? It just contains the setting and no code modification. Also how would this work with batching ? I'm assuming it's only when batching is not turned on ?

nlevitsky:
The code already contains this variable - ProfileServiceImpl - "forceRefreshOnSave", it was false by default and for some reason there was no environment variable that enables to set the value.

PR #171, opened 2020-06-26
UNOMI-356 Refactor REST API

sergehuber:
- A new RestServer class is responsible for creating a single JAX RS Server
- Service beans must register with the OSGi service registry and setup the following property: osgi.jaxrs.resource=true (this property comes from OSGi R7 so that we can then migrate to OSGi R6 JAX RS whiteboard once we are on OSGI R7)
- All JAX RS service beans are now register, not just the ones in the "rest" subproject
- Swagger UI is integrated by default, although we might make it optional in the future.
- Upgraded all OSGi JARs from version 4.3.1 to 6
- Remove usage of Blueprint in the "rest" subproject as it is no longer needed, it is replaced by OSGi DS annotations
- Removal of Miredot Maven plugin to generate REST API documentation as it is replaced by Swagger/OpenAPI generation.

Despite these changes the REST API is 100% compatible with the previous integration, which was by design a requirement.

To test the Swagger UI you can load it by going to http://localhost:8181/cxs/api-docs?url=openapi.json


rest/src/main/java/org/apache/unomi/rest/RestServer.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+@Component
+public class RestServer {
+
+    private static final Logger logger = LoggerFactory.getLogger(RestServer.class.getName());
+
+    private Server server;
+    private BundleContext bundleContext;
+    ServiceTracker jaxRSServiceTracker;
```
Taybou, line 51:
```suggestion
    private ServiceTracker jaxRSServiceTracker;
```

PR #172, opened 2020-06-29
UNOMI-357 Refresh index

giladw:


sergehuber:
Could you please create a JIRA ticket so that we can track this improvement? Otherwise it looks good to me.

giladw:
Hi,
opened the Jira issue, https://issues.apache.org/jira/browse/UNOMI-357
thanks for reviewing :)

PR #175, opened 2020-07-06
UNOMI-354 Add CDP_LongProperty input and output types

pmi:
NOTE: Java Long (max value 2^63-1) and JavaScript Number ( max value 2^53 - 1) are different, therefore full java long is not safe as stated here https://www.graphql-java.com/documentation/v14/schema/

### DIFF
--- a/graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
+++ b/graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.graphql.types.input.property;
+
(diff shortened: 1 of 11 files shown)


graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
```diff
@@ -0,0 +1,84 @@
...
+    public void updateType(final PropertyType type) {
+        if (type == null) {
+            return;
+        }
+        super.updateType(type);
+        type.setDefaultValue(defaultValue != null ? defaultValue.toString() : null);
+        if (minValue != null || maxValue != null) {
+            final Double from = minValue != null ? minValue.doubleValue() : null;
+            final Double to = maxValue != null && (minValue == null || maxValue >= minValue) ? maxValue.doubleValue() : null;
+            updateDefaultNumericRange(type, from, to);
```
Taybou, line 79: where is this function declared?

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            deleteDefaultNumericRange(type);
```
Taybou, line 81: where is this function declared?

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
pmi: This is declared in BaseCDPPropertyInput, which is the parent class for all property inputs.

It's a custom method to save min/max property values as unomi numeric range with pre-defined name *defaultRange* because unomi doesn't have dedicated fields for saving min/max values as of now.

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/types/input/property/CDPLongPropertyInput.java
pmi: Same here, comes from BaseCDPPropertyInput, which is the parent class for all property inputs.

pmi:
@sergehuber Good point ! Will add tests now.

pmi:
Added an IT for long property, fixed failing IT for segment and also the set "properties" data fetcher.

PR #178, opened 2020-07-21
Use profile id if the session id is null when checking if event already raised

liatiusim:
Since adding events with profile ID instead of session ID is now supported, it should be supported also when checking if event has already been raised (otherwise null pointer exception will be thrown)


sergehuber:
Hello,

Thanks for the fix, is it possible to add an integration test for this? 

Regards,
  Serge...

liatiusim:
> Pinging again for an integration test.
> 
> Thanks

Sure, done

PR #179, opened 2020-07-24
Improve scripting security

sergehuber:
- OGNL is now disabled by default as it wasn't really used much (possible to reactive through a system setting)
- A new condition sanitizer has been added in the ContextServlet to filter out any MVEL scripts (again not used much and can be reactivate using a system property)
- A new ExpressionFilter has been added that will use configurable (system property) regular expressions to filter out possible malicious expressions
- OGNL sandboxing has been improved
- MVEL default imports have been modified to prevent using system-level classes


wab/src/main/java/org/apache/unomi/web/ContextServlet.java
```diff
@@ -470,4 +472,80 @@ public void setPersonalizationService(PersonalizationService personalizationServ
...
+        }
+        List<PersonalizationService.PersonalizationRequest> result = new ArrayList<>();
+        for (PersonalizationService.PersonalizationRequest personalizationRequest : personalizations) {
+            List<PersonalizationService.PersonalizedContent> personalizedContents = sanitizePersonalizedContentObjects(personalizationRequest.getContents());
+            if (personalizedContents != null && personalizedContents.size() > 0) {
+                result.add(personalizationRequest);
+            }
+        }
+        return result;
+    }
```
jkevan, line 511: I understand that this fcts are filtering the elements, it could be great to have some loging here for debuging matter.
If someone want to undersand why his conditions are not executed so it could find answers in the logs

sergehuber: Yes indeed I did that in other places so I think it makes sense to add it here.

scripting/src/main/java/org/apache/unomi/scripting/MvelScriptExecutor.java
```diff
@@ -24,11 +24,20 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+/**
+ * MVEL script executor implementation
+ */
 public class MvelScriptExecutor implements ScriptExecutor {
 
+    private final static String INVALID_SCRIPT_MARKER = "Invalid String Marker";
```
Taybou, line 32: is it correct the placeholder text, or typo (script // string)?

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
```diff
@@ -305,10 +311,16 @@ protected Object getHardcodedPropertyValue(Item item, String expression) {
...
-                    String expressionPart = expression.substring("target.properties.".length());
-                    return getNestedPropertyValue(expressionPart, customItem.getProperties());
+                String expressionPart = expression.substring("target.properties.".length());
+                Item targetItem = event.getTarget();
+                if (targetItem instanceof CustomItem) {
+                    return getNestedPropertyValue(expressionPart, ((CustomItem) targetItem).getProperties());
+                } else if (targetItem instanceof Session) {
+                    return getNestedPropertyValue(expressionPart, ((Session) targetItem).getProperties());
+                } else if (targetItem instanceof Rule) {
+                    return null;
```
Taybou, line 321: is it okay to return null in case of Rule?

Taybou (approved):
Big work! 

Thank you @sergehuber

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
sergehuber: Yes should be ok, it's testing target.properties which is not available on the Rule class

scripting/src/main/java/org/apache/unomi/scripting/MvelScriptExecutor.java
sergehuber: It's a typo I'll correct it.

jkevan (approved):

PR #185, opened 2020-08-17
UNOMI-368 add propertyValueDouble to condition evaluator

Taybou:


sergehuber (approved):
Thanks for adding the integration tests, this looks good.

PR #186, opened 2020-08-20
Unomi - 387: Stop Unomi in case of an error which its cannot recovery from

liatiusim:
In case elastic reactor is stopped, unomi starts an unhealthy loop, which requires crashing the app.


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -2081,13 +2081,25 @@ public T executeInClassLoader(Object... args) throws Exception {
...
             try {
                 return executeInClassLoader(timerName, args);
-            } catch (Throwable t) {
-                if (logError) {
-                    logger.error("Error while executing in class loader", t);
+            } catch (IllegalStateException e) {
+                // Reactor stopped - recovery is not possible
+                if (e.getMessage().contains("I/O reactor status: STOPPED")) {
+                    logger.error("I/O Reactor stopped - stopping application");
+                    System.exit(-1);
```
sergehuber, line 2088: System.exit is really not recommended in an OSGi environment. It would be better to use the UnomiManagementService.stopUnomi method.

Ideally we would add a way for the ElasticSearch persistence implementation to recover from a failing connection but that's a lot more involved in terms of work.

sergehuber (changes requested):

sergehuber:
Hello, thank you for the PR, I put a comment in the code above.

sergehuber:
@liatiusim have you seen my comments on this PR ?

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
liatiusim: Thank you for the code review. I'll fix that soon

sergehuber:
Also please create a JIRA ticket for this PR and add it to the title of the PR. 

Thank you.

liatiusim:
Hey Serge, sorry for the late response.
I added a Jira ticket for this PR. As for your comment about using the bundleContext.close(), I fixed it (didn't pushed yet), but I think that since other threads are running while the system bundle context is closed, there are errors and the application is not gracefully shutting down. For example, I get this error - 
```
org.ops4j.pax.web.pax-web-extender-war [org.ops4j.pax.web.extender.war.internal.WebObserver] ERROR : Error scanning web bundle org.apache.unomi.wab [162]: null
java.lang.NullPointerException
```
and other errors as well, depends on the specific run.
Probably because other threads are running and the bundles were already closed.
Do you have any suggestions how to handle that?
Thanks

PR #188, opened 2020-08-31
Unomi 1.5.x throw exceptions

liatiusim:
Currently if there is an exception in Unomi, it is catched and 204 return in the response.
This PR allows to use the option of throwException=false/true in the custom.system.properties.
If this field is true, the response in case of exception would be an error.
Note that the PR is for unomi-1.5.x branch, since it changes files that are no longer exist in master branch


sergehuber:
@liatiusim do you have a version for the master ? Because I would like to avoid having something only in the 1.5 branch if it is not in master.

sergehuber:
@liatiusim  Also there really should be a ticket in UNOMI JIRA for these modifications. I forgot to ask before, but it will make tracking changes and releasing the changelog a lot easier. Could you please create a ticket there ? https://issues.apache.org/jira/projects/UNOMI/summary

PR #189, opened 2020-09-03
UNOMI-365 Create new Unomi Topic object & queries & mutations

anatol-sialitski:


sergehuber (approved):

sergehuber:
Thank you so much for implementing that and especially for the IT tests. 

Sorry for being so late on the review.

PR #192, opened 2020-09-11
UNOMI-370: fix device category resolution by using setter correctly.

jkevan:


Taybou (approved):
Thank you for the fix

PR #199, opened 2020-09-21
Allow a single index for all events (remove monthly index)

lyogev:
Allow a single index for all events (remove monthly index)


sergehuber (approved):
Thanks for the PR, looks good to me !

PR #201, opened 2020-09-24
UNOMI-366: Implement increment interest event type & action

anatol-sialitski:


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementInterestAction.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private Double interestsMinValue;
+
+    private Double interestsMaxValue;
+
+    private Double interestsDividerValue;
+
+    @Override
+    public int execute(final Action action, final Event event) {
+        final Map<String, Double> interestsAsMap = (Map<String, Double>) event.getProperty("interests");
```
sergehuber, line 48: In order to make this compatible with old rules, could we change this so that we use a parameter that says where we are retrieving the interests from ? This was we could say : interests or target.properties.interests like it was in old (legacy) page events. See UNOMi-366 for an example.

sergehuber (changes requested):
Thank you for the PR. I am requesting an adjustment to help with compatibility with legacy events.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementInterestAction.java
anatol-sialitski: I've updated my code. Please have a look

sergehuber: I'm sorry I wasn't clear enough. What I meant is that until now when a page view event was sent, it would also contain a list of topics that are metadata on the page and that should be used to increment the interests on the profile. Here's an example of such an event:

unomiTracker.initialize({
           ​scope: 'myScope',
           ​url: 'http://unomi:8181', // we use an empty URL to make it relative to this page.
           ​initialPageProperties: {
               ​path: path,
               ​pageInfo: {
                   ​destinationURL: location.href,
                   ​tags: ["tag1", "tag2", "tag3"],
                   ​categories: ["category1", "category2", "category3"]
               ​},
               ​interests: {
                   ​"interest1": 1,
                   ​"interest2": 2,
                   ​"interest3": 3
               ​}
           ​}
       ​})

I was hoping we could modify the action to also be able to extract the interests from this type of event, which is why I was suggesting to have an action parameter like "eventInterestProperty" : "target.properties.interests"

sergehuber (approved):
Thanks Anatol for changing the PR. It now looks good. 

I've approved it and will merge it now.

felipeelia:
Hey @sergehuber, is there a plan to merge this in any future release?

sergehuber:
Hi @felipeelia this is currently merged into the master so it is currently scheduled for Unomi 2. As for backports into 1.x I'm not sure. 

Actually I was looking at creating a more generic action that could increment any property.

PR #203, opened 2020-10-01
Implement CDP_View GET by performing aggregate query on scopes

anatol-sialitski:


sergehuber (approved):
Thank you for the PR, looks good to me

PR #204, opened 2020-10-05
Add toDate and fromDate to pastEventCondition

liatiusim:
Add option to use fromDate and toDate to pastEvent condition.
The logic of the time range would be an AND between fromDate/toDate/numberOfDays


sergehuber (approved):
Thanks for the PR, looks good to me. Thanks for including integration tests

PR #208, opened 2020-10-28
UNOMI-389: Add optimizations for past event queries

liatiusim:
Add configurable params for running optimized past event queries:
1. Option to disable partitions for aggregation queries
2. Option to set elasticsearch client socket timeout
3. Option to return error in case of missing docs in aggregation queries
4. Option to set the max http response size in elastic client for aggregation queries


sergehuber (approved):
Hello thanks for the PR, it looks good to me. 

It would be great if integration tests could be added in a separate PR.

PR #209, opened 2020-11-03
UNOMI-391: provide SameSite=Lax configuration on profile id cookie

jkevan:


sergehuber (approved):
Hello Kevan, 

Thank you for the PR. Just for next time, we usually do the PRs against the master and cherry pick them to the maintenance branches, so that the master includes all the fixes. 

Otherwise it looks good. 

cheers,
  Serge..

PR #211, opened 2020-11-13
UNOMI-399: fix the loading of MVEL allowed script pattern from config…

jkevan:
…, and return null during script execution if the script is filtered out


sergehuber (approved):
Thanks a lot for the PR, looks good to me will merge.

PR #213, opened 2020-11-16
UNOMI-399: load allow/forbid script from files instead of configurati…

jkevan:
…on property to avoid script code conflicting with property value parsing


sergehuber (approved):
Thanks a lot for the PR. Looks good to me.

PR #215, opened 2020-11-16
UNOMI-401: correctly set ClassLoader before MVEL script execute

jkevan:


sergehuber (approved):
Hi Kevan, thanks for doing this so fast and so late :) 

All good to me.

PR #217, opened 2020-11-17
UNOMI-399: Provide doc inside the custom.system.properties for script…

jkevan:


sergehuber (approved):
Thanks for the PR Kevan !

PR #218, opened 2020-11-17
UNOMI-400 Refactoring of hardcoded property accessors

sergehuber:
In this PR we externalize the hardcoded property accessors and make better use of polymorphism to make the code cleaner and easier to understand and maintain.


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/accessors/HardcodedPropertyAccessorRegistry.java
```diff
@@ -0,0 +1,131 @@
...
+        if (expression.startsWith("[\"")) {
+            int lookupNameBeginPos = "[\"".length();
+            int lookupNameEndPos = expression.indexOf("\"].", lookupNameBeginPos);
+            if (lookupNameEndPos > lookupNameBeginPos) {
+                nextTokens.propertyName = expression.substring(lookupNameBeginPos, lookupNameEndPos);
+                nextTokens.leftoverExpression = expression.substring(lookupNameEndPos+2);
+            } else {
+                nextTokens.propertyName = expression.substring(lookupNameBeginPos);
+                nextTokens.leftoverExpression = null;
+            }
```
jkevan, line 70: `["foo"].bar` is covered
`["foo"]` is not covered

Also could be great to cover this function with unit test, since it's covering a lot of cases, and it should not be that complicated todo.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/accessors/HardcodedPropertyAccessorRegistry.java
```diff
@@ -0,0 +1,131 @@
...
+            int lookupNameBeginPos = ".".length();
+            int lookupNameEndPos = lookupNameBeginPos;
+            int dotlookupNameEndPos = expression.indexOf(".", lookupNameBeginPos);
+            int squareBracketlookupNameEndPos = expression.indexOf("[", lookupNameBeginPos);
+            if (dotlookupNameEndPos > lookupNameBeginPos && squareBracketlookupNameEndPos > lookupNameBeginPos) {
+                lookupNameEndPos = Math.min(dotlookupNameEndPos, squareBracketlookupNameEndPos);
+            } else if (dotlookupNameEndPos > lookupNameBeginPos) {
+                lookupNameEndPos = dotlookupNameEndPos;
+            } else if (squareBracketlookupNameEndPos > lookupNameBeginPos) {
+                lookupNameEndPos = squareBracketlookupNameEndPos;
```
jkevan, line 81: the two else if seem's unecessary if we do:
`dotlookupNameEndPos > lookupNameBeginPos || squareBracketlookupNameEndPos > lookupNameBeginPos`
in the first condition

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/accessors/HardcodedPropertyAccessorRegistry.java
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+            int lookupNameBeginPos = 0;
+            int lookupNameEndPos = expression.indexOf(".", lookupNameBeginPos);
+            if (lookupNameEndPos > lookupNameBeginPos) {
+                nextTokens.propertyName = expression.substring(lookupNameBeginPos, lookupNameEndPos);
+                nextTokens.leftoverExpression = expression.substring(lookupNameEndPos);
+            } else {
+                nextTokens.propertyName = expression.substring(lookupNameBeginPos);
+                nextTokens.leftoverExpression = null;
+            }
```
jkevan, line 101: this code is duplicated in the 3 cases, we could move it out of the tree cases and just calculated the `lookupNameBeginPos, lookupNameEndPos.`
Then do the this code at the end of the 3 cases check.

jkevan (changes requested):
Thank you for the PR, it's way much cleaner like that. 
And I like the Generics to avoid too much casting.

I think there is a missing case if expression is `["foo"]`.
Take a look at the first comment.

Also I propose 2 code improvements

sergehuber:
Ok thanks for the feedback Kevan. I've updated this branch with fixes for all the points you raised. Thanks for checking them out.

jkevan (approved):

PR #222, opened 2020-11-29
Fix bug in profile service it

liatiusim:


sergehuber (approved):
Hi Liat,

This is a very interesting way to implement this, I didn't think of using Reflection API. 

Thanks for doing this, I will merge it.

Regards,
  Serge..

PR #223, opened 2020-12-02
UNOMI-371 add optional support for optimistic concurrency control (if…

giladw:
…_seq_no)


sergehuber:
Thanks for the contribution it is now merged. I'll be adding a small fixed I found at review.

PR #228, opened 2020-12-17
UNOMI-409 add .gitignore to unomi-web-tracker

Taybou:


extensions/web-tracker/javascript/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
sergehuber, line 1: I think we should ignore the snippet.min.js and the snippet.min.js.map instead of committing them WDYT ?

sergehuber (changes requested):

extensions/web-tracker/javascript/.gitignore
Taybou: yes, in this case, we can move snippet.min.js && snippet.min.js.map into dist, is it okay?

sergehuber: I guess so, I'm not an expert at NPM packaging, would that make sense from that point of view?

Taybou: yes, I'll handle it, thanks

sergehuber (approved):

sergehuber:
Thanks for the changes, looks good to me will merge.

PR #229, opened 2020-12-19
UNOMI-411 Bug fix in profile property condition and add getNestedProperty method

liatiusim:
When querying using "exist" operator in Condition, an existQuery is created and sent to Elasticsearch. According to ES documentation, exist return false the the field in null or empty list - "[]".

The case of an empty list wasn't handled in the eval() method of profile property condition evaluator. Need to return false if the field exist but has a value which is empty list.

### DIFF
--- a/api/src/main/java/org/apache/unomi/api/Profile.java
+++ b/api/src/main/java/org/apache/unomi/api/Profile.java
@@ -102,6 +102,10 @@ public class Profile extends Item {
      * @return the value of the property identified by the specified name
      */
     public Object getNestedProperty(String name) {
+        if (!name.contains(".")) {
+            return getProperty(name);
+        }
+
         Map properties = this.properties;
         String[] propertyPath = StringUtils.substringBeforeLast(name, ".").split("\\.");
         String propertyName = StringUtils.substringAfterLast(name, ".");
--- a/api/src/main/java/org/apache/unomi/api/Profile.java
+++ b/api/src/main/java/org/apache/unomi/api/Profile.java
@@ -17,6 +17,7 @@
 
 package org.apache.unomi.api;
 
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/apache/unomi/api/Profile.java
```diff
@@ -94,6 +95,30 @@ public Object getProperty(String name) {
...
     }
 
+    /**
+     * Retrieves the value of the nested property identified by the specified name.
+     *
+     * @param name the name of the property to be retrieved, splited in the nested properties with "."
+     * @return the value of the property identified by the specified name
+     */
+    public Object getNestedProperty(String name) {
+        if (!name.contains(".")) {
```
sergehuber, line 105: I'm wondering if there is a way we could avoid this code duplication but I don't know where to put it unless we create some kind of static class in the API.

sergehuber (changes requested):
Thanks for the PR. I put a note about code duplication. If we could find a better way it would be preferable I think.

sergehuber:
Could you also implement an integration test for this?

api/src/main/java/org/apache/unomi/api/Profile.java
liatiusim: Thank you for the comment. I'm not sure if I know where to put the common code, what do you think?

liatiusim:
Added integration tests

api/src/main/java/org/apache/unomi/api/Profile.java
sergehuber: Yeah let's leave it like this for the time being but we'll probably have to refactor at some point.

sergehuber (approved):
Thanks for adding the IT test and for your patience. Looks good to me now I'll approve and merge.

PR #230, opened 2020-12-29
feat(updatePropertiesAction): support add values to a SET in a profil…

giladw:
…e property


sergehuber (approved):
Thanks for the contribution. Looks good to me, will merge.

PR #231, opened 2020-12-30
UNOMI-410 : add patch template

jsinovassin:
### Description

The patch.json is added with this PR to allow to create the indexes in elastic search for the patch index, this taking into account the configuration file.

### JIRA issue
https://issues.apache.org/jira/browse/UNOMI-410


Taybou (approved):
Thank you for the fix

sergehuber (approved):
Looks good to me thanks for the contribution. Will merge.

PR #232, opened 2021-01-03
UNOMI-413-mergeProfileOnPropertyAction: support take Events profile as ma…

giladw:
…ster, fix persistence bug when new profile is merged and it was the event profile


sergehuber:
Thanks for the contribution, it is now merged

PR #235, opened 2021-01-13
UNOMI-414 Fix the ScorePersonalizationStrategy

romain-pm:
Avoid nullpointer when the profile doesnt have any interests


sergehuber (changes requested):
Thanks for the contribution, it looks good but could you please do it against the master branch, as we cherry pick to stable branches usually ?

sergehuber:
I'll implement an integration test and merge this to both branches as soon as finished.

PR #237, opened 2021-01-19
UNOMI-421 - update segments with its own bulk request, and use retry for failures

giladw:


giladw:
https://issues.apache.org/jira/browse/UNOMI-421

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -235,12 +237,13 @@
     private Set<String> itemClassesToCacheSet = new HashSet<>();
     private String itemClassesToCache;
     private boolean useBatchingForSave = false;
+    private boolean useBatchingForUpdate = true;
+    private boolean alwaysOverwrite = true;
+    private boolean refreshBeforeQuery = false;
```
jkevan, line 242: This prop seem's not used.

persistence-spi/src/main/java/org/apache/unomi/persistence/spi/PersistenceService.java
```diff
@@ -175,6 +176,17 @@
...
+    /**
+     * Updates Map of items of the specified class and identified by the specified identifier with a new property value for the specified property name. Same as
+     * {@code update(itemId, dateHint, clazz, Collections.singletonMap(propertyName, propertyValue))}
+     *
+     * @param items         A map the consist of item (key) and properties to update (value)
+     * @param dateHint      a Date helping in identifying where the item is located
+     * @param clazz         the Item subclass of the item to update
+     * @return List of failed Items Ids, if all succesful then returns an empty list. if the whole operation failed then will return null
+     */
+    List<String> update(Map<Item, Map> items, Date dateHint, Class clazz);
```
jkevan, line 188: It's strange to have a Map<Item, ...> this mean that it could contains items of different types, and it may not correspond to the Class clazz parameter.
Could we implement a check before persisting the item that the item is correctly of type clazz ?

services/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -34,7 +34,11 @@
             <cm:property name="definitions.refresh.interval" value="10000"/>
             <cm:property name="properties.refresh.interval" value="10000"/>
             <cm:property name="segment.refresh.interval" value="1000"/>
+            <cm:property name="segment.max.retries.update.profile.segment" value="5"/>
+            <cm:property name="segment.retry.update.segment.seconds.delay" value="1"/>
```
jkevan, line 38: This one is not listed in the configurable values, is it normal ?

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
```diff
@@ -919,74 +943,93 @@ private void updateExistingProfilesForSegment(Segment segment) {
...
-                if (profilesToRemove == null || profilesToRemove.getList().size() == 0) {
-                    break;
+            else { //send update profile one by one
+                for (Profile profileToUpdate : profiles.getList()) {
+                    Map<String, Object> sourceMap = buildPropertiesMapForUpdateSegment(profileToUpdate, segmentId, isAdd);
+                    persistenceService.update(profileToUpdate, null, Profile.class, sourceMap);
                 }
             }
+           if (sendProfileUpdateEventForSegmentUpdate)
+                sendProfileUpdatedEvent(profiles.getList());
```
jkevan, line 970: In case of batch and failure, this will still send an event if the profile failed to be updated, seem's wrong.
We should send the event only if we are sure the update worked.

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
```diff
@@ -919,74 +943,93 @@ private void updateExistingProfilesForSegment(Segment segment) {
...
-                    eventService.send(profileUpdated);
-                    updatedProfileCount++;
-                }
-                logger.info("{} profiles removed from segment in {}ms", profilesToRemove.size(), System.currentTimeMillis() - profilesToRemoveStartTime);
-                profilesToRemove = persistenceService.continueScrollQuery(Profile.class, profilesToRemove.getScrollIdentifier(), profilesToRemove.getScrollTimeValidity());
-                if (profilesToRemove == null || profilesToRemove.getList().size() == 0) {
-                    break;
-                }
-            }
+            updatedProfileCount += profiles.size();
```
jkevan, line 972: same for the count, it will also contains profiles that failed.
We should only count the profiles that update successfully.

jkevan (changes requested):

services/pom.xml
```diff
@@ -74,6 +74,18 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>net.jodah</groupId>
+            <artifactId>failsafe</artifactId>
+            <version>2.4.0</version>
+        </dependency>
```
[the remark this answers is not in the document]
giladw: removed

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
giladw: removed "refreshBeforeQuery"

persistence-spi/src/main/java/org/apache/unomi/persistence/spi/PersistenceService.java
giladw: i wonder where to have it and how.
the implementation of this method does  use the class for every item request in the bulk request.
also regular update uses a class and item without checks, they rely that it will on elastic if index is not found.

https://github.com/apache/unomi/pull/237/files#diff-cb0ba8419f934aeaa557cb1eded52b588ae98b8f1b9dbe6abd1d67553a0b27b2R977

https://github.com/apache/unomi/pull/237/files#diff-cb0ba8419f934aeaa557cb1eded52b588ae98b8f1b9dbe6abd1d67553a0b27b2R949

services/src/main/resources/OSGI-INF/blueprint/blueprint.xml
giladw: right, added it for "segment.retry.update.segment.seconds.delay" as well

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
giladw: we dont have this check for regular update also (not batch).
maybe it should be done separately?
(currently there is also no handling of errors)

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
giladw: see comment above.
also the code assumes currently that the updates works and if not at first I added a retry.
but there is still no handling of errors and exceptions

persistence-spi/src/main/java/org/apache/unomi/persistence/spi/PersistenceService.java
jkevan: Yes you are right even the regular single update is doing that, but I think it would be great to do the same check.

before processing the item we could check for:
`item.getClass().isAssignableFrom(clazz.class)`
if not
`throw new IllegalArgumentException();`

But it could be done and check in a separate PR/unomi ticket, as it is not the current PR topic.

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
jkevan: We have the List<String> failedItemsIds as the return from persistenceService.update.
So we could filter and only send the event for the not failed update and update the count accordingly.
But as it was already the case before, it can also be part of a separate improvement.

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
jkevan: since the retry is synchron we could also get the failed ids from the retry, and send the events for only successful updated events, and udpate the count accordingly.
In term of implementation, event it was the case before it's sounds not correct to considered a failed update profile as counted in the number of profiled updated, and also send the updated event for a profile that was not updated.
But again, since can be done in a separate unomi PR and ticket.

jkevan (approved):

PR #243, opened 2021-02-04
UNOMI-424 Fix buildQuery for BooleanConditionESQueryBuilder

Taybou:
add and fix IT


sergehuber (approved):
Looks good to me thanks for the contribution. I will merge it now.

PR #245, opened 2021-02-04
UNOMI-425 Fix IT

Taybou:


sergehuber (approved):
Thanks for finding and fixing this ! 

Looks good to me will merge.

PR #247, opened 2021-02-08
UNOMI-428: Configure max profiles to merge instead use max size

liatiusim:
JIRA ticket - https://issues.apache.org/jira/browse/UNOMI-428
Added the ability to configure max profiles to merge in a merge action (instead using "-1" hard coded), due to performance issues - when the size is "-1" , the `query()` call generates additional 2 calls - scroll query and the deletion of the scroll.
This is not needed when we know there is a few profiles to merge.

note - since the baseplugin has no cgf file, I wasn't sure what should be done in order to expose it to environment variable (currently it is configurable only from the blueprint file). Would be glad to get your advise about that. Thanks


sergehuber (changes requested):
Hello Liat, thanks for the PR, but it doesn't seem complete, I don't see where the new variable is being used?

sergehuber:
To address the question of where to put the configuration, a new configuration file for the base plugin could be added, something like org.apache.unomi.plugins.base.cfg. See the other bundles for examples on how to set that up.

sergehuber:
Hi @liatiusim any updates on this PR?

liatiusim:
Hey Serge,
Added the Jira number and added a cfg file
Thanks for the review

sergehuber (approved):
Thanks Liat for providing all the files. Looks good to me.

PR #248, opened 2021-02-08
UNOMI-426 remove unneeded stages in Jenkinsfile-jdk11

Taybou:


sergehuber (approved):
Thanks for this Taybou, looks good to me.

PR #250, opened 2021-02-15
UNOMI-431 Add initial proposal for PR template

sergehuber:
Currently we have no consistent pull request structure, although we do have processes that we'd like to have contributors follow.

The idea is to use Github's Pull Request Template feature to pre-populate each PR with a list of things to remember to do.

Please let me know what you think of it and any suggestions for improvement you might have.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -30,4 +30,3 @@ To make clear that you license your contribution under the [Apache License Versi
 you have to acknowledge this by using the following check-box.
 
  - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+If this is your first contribution, you have to read the [Contribution Guidelines](https://unomi.apache.org/contribute.html)
+
+If your pull request is about ~20 lines of code you don't need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) 
+if you are unsure please ask on the developers list.
+
+To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
+you have to acknowledge this by using the following check-box.
+
+ - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
+ - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).
```
Taybou, line 33: why are there two checkboxes, maybe we should only keep one, wdyt?

Taybou (approved):
Looks good to me, thank you @sergehuber for the template ^^'

.github/pull_request_template.md
sergehuber: Yes I think you're right I'll simplify this part.

PR #252, opened 2021-02-18
UNOMI-350 Implement CDP_Source definition & retrieval

anatol-sialitski:
In scope this PR was implemented the following things:
- Deprecated `scope` for event object
- Added SourceItem object and related services with it
- `getSources` query; `createOrUpdateSource` and `deleteSource`  mutations of the GraphQL API
- Added a new system property UNOMI_SHOULD_BE_CHECKED_EVENT_SOURCE_ID
- Added IT tests


api/src/main/java/org/apache/unomi/api/Event.java
```diff
@@ -369,10 +372,12 @@ public void setProperties(Map<String, Object> properties) {
     /**
      * @return the scope
      */
+    @Deprecated
```
Taybou, line 375: why **@Deprecated** the scope of the event?

anatol-sialitski: According to the [ticket](https://issues.apache.org/jira/browse/UNOMI-350)

Taybou: Great, thank you

sergehuber (approved):
Thanks a lot for this contribution. It looks mostly good, but we'll have to set how it behaves with existing data sets. I'll merge it so that we can start testing.

PR #253, opened 2021-02-19
UNOMI-438 Fix broken GraphQL Playground

sergehuber:
- Remove chunking optimization from Webpack configuration
- Switch Webpack back to development mode.

### DIFF
--- a/graphql/graphql-playground/pom.xml
+++ b/graphql/graphql-playground/pom.xml
@@ -30,7 +30,7 @@
     <packaging>bundle</packaging>
 
     <properties>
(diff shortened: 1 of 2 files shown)


graphql/graphql-playground/pom.xml
```diff
@@ -30,7 +30,7 @@
     <packaging>bundle</packaging>
 
     <properties>
-        <yarn.arguments>build:production</yarn.arguments>
+        <yarn.arguments>build</yarn.arguments>
```
Taybou, line 33: we can keep build:production, to optimize the build at least

Taybou (approved):

PR #256, opened 2021-02-24
UNOMI-419: fix SegmentsIT test to wait on full startup

jkevan:


dgriffon (approved):

itests/src/test/java/org/apache/unomi/itests/SegmentIT.java
```diff
@@ -46,8 +47,15 @@
     @Filter(timeout = 600000)
     protected SegmentService segmentService;
 
+    @Inject @Filter(timeout = 600000)
+    protected BundleWatcher bundleWatcher;
+
     @Before
     public void setUp() throws InterruptedException {
+        while (!bundleWatcher.isStartupComplete()) {
```
sergehuber, line 55: Ideally we should move this setup to BaseIT since all the tests should have it, but we should do this in a separate PR.

PR #258, opened 2021-02-26
UNOMI-440: Add explicit call to update the current profile id of the event

liatiusim:
Jira task - https://issues.apache.org/jira/browse/UNOMI-440

On merge action, the profile id of the events is updated to be the master profile.
Since the event is saved in ES and right after is updated - it might not return in the query. As a result, the profile id might not be updated.

Added an explicit update.


sergehuber:
Hi @liatiusim thanks for this PR and sorry for taking so much time to review it. Looks good to me I'll merge it.

PR #260, opened 2021-03-10
UNOMI-442: fix creation of Segment that contains past event condition…

jkevan:
…, correctly engage profiles during creation process

### DIFF
--- a/services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
+++ b/services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
@@ -776,7 +776,7 @@ public class SegmentServiceImpl extends AbstractServiceImpl implements SegmentSe
                     rule.setLinkedItems(Arrays.asList(metadata.getId()));
                     rules.add(rule);
 
-                    updateExistingProfilesForPastEventCondition(condition, parentCondition);
+                    updateExistingProfilesForPastEventCondition(condition, parentCondition, true);
                 } else {
                     rule.getLinkedItems().add(metadata.getId());
                     rules.add(rule);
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
```diff
@@ -839,20 +839,25 @@ private void updateExistingProfilesForPastEventCondition(Condition eventConditio
...
             int numParts = (int) (card / aggregateQueryBucketSize) + 2;
             for (int i = 0; i < numParts; i++) {
                 Map<String, Long> eventCountByProfile = persistenceService.aggregateWithOptimizedQuery(andCondition, new TermsAggregate("profileId", i, numParts), Event.ITEM_TYPE);
-                updateProfilesWithPastEventProperty(eventCountByProfile, propertyKey);
+                updatedProfileCount += updateProfilesWithPastEventProperty(eventCountByProfile, propertyKey);
             }
         }
 
-        logger.info("Profiles past condition updated in {}ms", System.currentTimeMillis() - t);
+        if (forceRefresh && updatedProfileCount > 0) {
```
giladw, line 856: just to make sure, 
if the update is done with bulkProcessor, how do you know that the refresh happened after the bulk processor has finished?

jkevan: That the point of the PR, I'm not using the bulk processor anymore.
But to answer: you cannot be sure, even if you do a bulkProcessor.flush()
The bulkProcessor is doing the update asynchronously in a separate thread.
I did debug this part with logs and the bulkProcessor listener, I was able to see that the refresh always come before bulkProcessor perform the request.

For the current PR, I'm now using this function to do the update (that do not use the bulk processor): https://github.com/apache/unomi/blob/master/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java#L967-L997
Especially because I cannot handle the bulk processor asynchronous behavior, and I need the past event props to be saved before the segment save try to update the profiles.

PR #265, opened 2021-03-18
UNOMI-446 improve increment action and make it generic

Taybou:
----

**Please** following this checklist to help us incorporate your contribution quickly and easily:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/UNOMI) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
 - [X] Format the pull request title like `[UNOMI-XXX] - Title of the pull request`
 - [x] Provide integration tests for your changes, especially if you are changing the behavior of existing code or adding
       significant new parts of code.
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why. 
       Copy the description to the related JIRA issue
 - [X] Run `mvn clean install -P integration-tests` to make sure basic checks pass. A more thorough check will be 
        performed on your pull request automatically.
 
Trivial changes like typos do not require a JIRA issue (javadoc, project build changes, small doc changes, comments...). 
 
If this is your first contribution, you have to read the [Contribution Guidelines](https://unomi.apache.org/contribute.html)

If your pull request is about ~20 lines of code you don't need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) 
if you are unsure please ask on the developers list.

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

### DIFF
--- a/itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
+++ b/itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
@@ -1,0 +1,335 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License
+ */
+package org.apache.unomi.itests;
+
+import java.util.*;
+
+import javax.inject.Inject;
+
+import org.apache.unomi.api.*;
+import org.apache.unomi.api.actions.Action;
+import org.apache.unomi.api.conditions.Condition;
+import org.apache.unomi.api.rules.Rule;
+import org.apache.unomi.api.services.DefinitionsService;
+import org.apache.unomi.api.services.EventService;
+import org.apache.unomi.api.services.ProfileService;
+import org.apache.unomi.api.services.RulesService;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.ops4j.pax.exam.junit.PaxExam;
+import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
+import org.ops4j.pax.exam.spi.reactors.PerSuite;
+import org.ops4j.pax.exam.util.Filter;
+
+import static org.apache.unomi.itests.BasicIT.ITEM_TYPE_PAGE;
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerSuite.class)
+public class IncrementPropertyIT extends BaseIT {
+    private Profile profile;
+    private Rule rule;
+
+    @Inject
+    @Filter(timeout = 600000)
+    protected ProfileService profileService;
+
+    @Inject
+    @Filter(timeout = 600000)
+    protected EventService eventService;
+
+    @Inject
+    @Filter(timeout = 600000)
+    protected RulesService rulesService;
+
+    @Inject
+    @Filter(timeout = 600000)
+    protected DefinitionsService definitionsService;
+
+    @Before
+    public void setup() throws Exception {
+        profile = createProfile();
+        rule = new Rule();
+    }
+
+    @After
+    public void tearDown() {
+        rulesService.removeRule(rule.getItemId());
+        profileService.delete(profile.getItemId(), false);
+    }
+
+    @Test
+    public void testIncrementNotExistingPropertyByAction() throws InterruptedException {
+        Action incrementPropertyAction = new Action(definitionsService.getActionType("incrementPropertyAction"));
+        incrementPropertyAction.setParameter("propertyName", "pageView.acme");
+
+        createRule(incrementPropertyAction);
+
+        CustomItem target = new CustomItem("ITEM_ID_PAGE", ITEM_TYPE_PAGE);
+        target.setScope("acme-space");
+
+        Event event = new Event("view", null, profile, null, null, target, new Date());
+        event.setPersistent(false);
+
+        int eventCode = eventService.send(event);
+        refreshPersistence();
+
+        if (eventCode == EventService.PROFILE_UPDATED) {
+            Profile updatedProfile = profileService.save(event.getProfile());
+            refreshPersistence();
+
+            int value = ((Map<String, Integer>) updatedProfile.getProperty("pageView")).get("acme");
+            Assert.assertEquals(1, value, 0.0);
+        } else {
+            throw new IllegalStateException("Profile was not updated");
+        }
+    }
+
+    @Test
+    public void testIncrementExistingPropertyByAction() throws InterruptedException {
+        Action incrementPropertyAction = new Action(definitionsService.getActionType("incrementPropertyAction"));
+        incrementPropertyAction.setParameter("propertyName", "pageView.acme");
+
+        createRule(incrementPropertyAction);
+
+        Map<String, Object> properties = new HashMap<>();
+        Map<String, Integer> propertyValue = new HashMap<>();
+        propertyValue.put("acme", 49);
+        properties.put("pageView", propertyValue);
+        profile.setProperties(properties);
+
+        CustomItem target = new CustomItem("ITEM_ID_PAGE", ITEM_TYPE_PAGE);
+        target.setScope("acme-space");
+
+        Event event = new Event("view", null, profile, null, null, target, new Date());
+        event.setPersistent(false);
+
+        int eventCode = eventService.send(event);
+        refreshPersistence();
+
+        if (eventCode == EventService.PROFILE_UPDATED) {
+            Profile updatedProfile = profileService.save(event.getProfile());
+
+            refreshPersistence();
+
+            int value = ((Map<String, Integer>) updatedProfile.getProperty("pageView")).get("acme");
+            Assert.assertEquals(50, value, 0.0);
+        } else {
+            throw new IllegalStateException("Profile was not updated");
+        }
+    }
+
+    @Test
+    public void testIncrementExistingPropertyWithExistingEventPropertyByAction() throws InterruptedException {
+        Action incrementPropertyAction = new Action(definitionsService.getActionType("incrementPropertyAction"));
+        incrementPropertyAction.setParameter("propertyName", "pageView.acme");
+        incrementPropertyAction.setParameter("propertyTarget", "project.nasa");
+
+        createRule(incrementPropertyAction);
+
+        Map<String, Object> properties = new HashMap<>();
+        Map<String, Integer> propertyValue = new HashMap<>();
+        propertyValue.put("acme", 49);
+        properties.put("pageView", propertyValue);
+        profile.setProperties(properties);
+
+        CustomItem target = new CustomItem("ITEM_ID_PAGE", ITEM_TYPE_PAGE);
+        target.setScope("acme-space");
+        properties = new HashMap<>();
+        propertyValue = new HashMap<>();
+        propertyValue.put("nasa", 19);
+        properties.put("project", propertyValue);
+        target.setProperties(properties);
+
+        Event event = new Event("view", null, profile, null, null, target, new Date());
+        event.setPersistent(false);
+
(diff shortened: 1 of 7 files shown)


itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
```diff
@@ -0,0 +1,305 @@
...
+            int value = ((Map<String, Integer>) updatedProfile.getProperty("pageView")).get("acme");
+            Assert.assertEquals(1, value, 0.0);
+        } else {
+            throw new IllegalStateException("Profile was not updated");
+        }
+    }
+
+    @Test
+    public void testIncrementExistingPropertyByAction() throws InterruptedException {
+        Action incrementPropertyAction = new Action(definitionsService.getActionType("incrementPropertyAction"));
```
sergehuber, line 207: Could you please add a test for the case of the ${} marker?

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -16,56 +16,88 @@
...
     @SuppressWarnings("unchecked")
     public int execute(final Action action, final Event event) {
-        Map<String, Double> interestsAsMap = (Map<String, Double>) action.getParameterValues().get( ACTION_INTERESTS_PROPERTY );
+        final Profile profile = event.getProfile();
 
-        if ( interestsAsMap == null ) {
-            interestsAsMap = (Map<String, Double>) event.getProperty( EVENT_INTERESTS_PROPERTY );
+        Object propertyValue;
+        String propertyName = (String) action.getParameterValues().get("propertyName");
+        String parentPropertyName = propertyName.split("\\.")[0];
```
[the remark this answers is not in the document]
sergehuber: Maybe it's that's normal, maybe we could rename it to rootPropertyName?

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -16,56 +16,88 @@
...
 
-            if (interestsAsMap == null) {
-                return EventService.NO_CHANGE;
+        try {
+            if (EVENT_INTERESTS_PROPERTY.equals(propertyName)) {
+                propertyValue = incrementInterests(action, event, profile);
```
sergehuber, line 59: I think we should rename this method to processIncrementInterestEvent

sergehuber: After reflection I wonder if we should split this action into 2 separate actions: IncrementInterests and IncrementProperty, because their logic is quite different.

services/src/main/java/org/apache/unomi/services/actions/ActionExecutorDispatcher.java
```diff
@@ -124,13 +126,21 @@ public Action getContextualAction(Action action, Event event) {
...
-                    // check if we have special values
-                    if (s.contains(VALUE_NAME_SEPARATOR)) {
-                        final String valueType = StringUtils.substringBefore(s, VALUE_NAME_SEPARATOR);
-                        final String valueAsString = StringUtils.substringAfter(s, VALUE_NAME_SEPARATOR);
-                        final ValueExtractor extractor = valueExtractors.get(valueType);
-                        if (extractor != null) {
-                            value = extractor.extract(valueAsString, event);
+                    if (s.contains(PLACEHOLDER_PREFIX)) {
+                        while (s.contains(PLACEHOLDER_PREFIX)) {
+                            String substring = s.substring(s.indexOf(PLACEHOLDER_PREFIX) + 2, s.indexOf(PLACEHOLDER_SUFFIX));
```
sergehuber, line 131: What happens if the PLACEHOLDER_SUFFIX is not found, does it raise an exception because indexOf is < 0 ?

sergehuber (changes requested):
Hello Taybou, I found a few issues/improvements that I put directly in the comments. Thanks.

services/src/main/java/org/apache/unomi/services/actions/ActionExecutorDispatcher.java
```diff
@@ -146,6 +156,19 @@ public Action getContextualAction(Action action, Event event) {
...
+        Object value = null;
+
+        String valueType = StringUtils.substringBefore(s, VALUE_NAME_SEPARATOR);
+        String valueAsString = StringUtils.substringAfter(s, VALUE_NAME_SEPARATOR);
+        ValueExtractor extractor = valueExtractors.get(valueType);
+        if (extractor != null) {
+            value = extractor.extract(valueAsString, event);
+        }
+
+        return value;
```
sergehuber, line 169: We should check the value to make sure it doesn't have any PLACEHOLDER_PREFIX in which case we might need to encode it somehow, otherwise this could lead to a security vulnerability as attackers could inject expressions that could ready any data.

sergehuber (changes requested):

jkevan:
Small update on what we discussed about in the call:

- Do not touch the unomi IncrementInterestAction. This action have his own logic and cannot be generify with property incrementation, also the interest incrementation from JExperience is not using this action, it's actually using a script and the setPropertyAction.
- Create a new Action: IncrementPropertyAction: 
  - propertyName: to be able to increment or create a single prop (properties.pageViewCount)
  - rootPropertyName: to be able to increment all props under this one: (ex: profile.properties.interest)
  - incrementWithValues: to be able to increment properties with the values coming from the event. (ex: event.properties.interest)
  - storeInSession: to be able to target the profile or the session.

Real exemple coming from JEXperience:
the increment page count: 

`"actions": [
    {
      "parameterValues": {
        "setPropertyName": "properties.pageViewCount",
        "setPropertyValue": "script::r = profile.properties['pageViewCount']; if (r == null) { profile.properties['pageViewCount'] = []; profile.properties.pageViewCount = [event.scope : 1] } else { if (r[event.scope] != null) { r[event.scope] = r[event.scope] + 1 } else { r[event.scope] = 1 }} r",
        "storeInSession": false
      },
      "type": "setPropertyAction"
    }
  ]`

We should be able to do it with this kind of conf:
type: incrementPropertyAction
paramerterValues:
- propertyName: properties.pageViewCount
- storeInSession: false

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -0,0 +1,113 @@
...
+        Object propertyValue = 1;
+
+        try {
+            if (StringUtils.isNotEmpty(propertyTarget)) {
+                value = PropertyUtils.getNestedProperty(((CustomItem) event.getTarget()).getProperties(), propertyTarget);
+            }
+
+            if (value != null) {
+                if (value instanceof Integer) {
+                    if (profile.getProperty(rootPropertyName) != null) {
```
jkevan, line 61: What happen in case
propertyName = root.foo.bar
propertyTarget = bar
profile.root = {blabla:blabla}
event.bar = 15

the getNestedProperty will fail because the rootPropertyName will be "root" and the profile have the "root" property, but doesnt have the "root.foo".
To be verified, but I found the calculation and the usage of the "rootPropertyName" not clear, and hard to understand.

At the end it seem's only use to see if the profile have the property.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -0,0 +1,113 @@
...
+        boolean storeInSession = Boolean.TRUE.equals(action.getParameterValues().get("storeInSession"));
+        if (storeInSession && event.getSession() == null) {
+            return EventService.NO_CHANGE;
+        }
+
+        Profile profile = event.getProfile();
+        String propertyName = (String) action.getParameterValues().get("propertyName");
+        String propertyTarget = (String) action.getParameterValues().get("propertyTarget");
+        String rootPropertyName = propertyName.split("\\.")[0];
+        Object value = null;
```
jkevan, line 51: Provide better name for the variable for better understanding the code could be great, this is the target value from the event

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -0,0 +1,114 @@
...
+        Object propertyTargetValue = null;
+
+        if (StringUtils.isNotEmpty(propertyTarget)) {
+            propertyTargetValue = PropertyUtils.getNestedProperty(((CustomItem) event.getTarget()).getProperties(), propertyTarget);
+        }
+
+        if (propertyTargetValue != null) {
+            if (propertyTargetValue instanceof Integer) {
+                if (properties.containsKey(rootPropertyName)) {
+                    propertyValue = (int) propertyTargetValue + (int) PropertyUtils.getNestedProperty(properties, propertyName);
```
jkevan, line 78: Careful with cast: (int) PropertyUtils.getNestedProperty(properties, propertyName);
- the PropertyUtils.getNestedProperty(properties, propertyName); can return null ?
- the PropertyUtils.getNestedProperty(properties, propertyName); can return a String, if the prop exist and is a String

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
```diff
@@ -0,0 +1,114 @@
...
+                } else {
+                    propertyValue = propertyTargetValue;
+                }
+            } else if (propertyTargetValue instanceof Map) {
+                if (properties.containsKey(rootPropertyName)) {
+                    Map<String, Object> nestedProperty = (Map<String, Object>) PropertyUtils.getNestedProperty(properties, propertyName);
```
jkevan, line 84: We should check that PropertyUtils.getNestedProperty(properties, propertyName); is actually not null and it's a map, if not log an error and continue.

itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
```diff
@@ -0,0 +1,411 @@
...
+        actions.add(incrementPropertyAction);
+
+        rule.setCondition(condition);
+        rule.setActions(actions);
+        rule.setMetadata(metadata);
+        rulesService.setRule(rule);
+        refreshPersistence();
+    }
+
+    private int buildAction(String propertyName, String propertyTargetName, Map<String, Object> properties, Map<String, Object> targetProperties) throws InterruptedException {
```
jkevan, line 364: Would rename "buildActionAndSendEvent"

itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
```diff
@@ -0,0 +1,411 @@
...
+        int eventCode = buildAction("pageView.acme", null, null, null);
+
+        if (eventCode == EventService.PROFILE_UPDATED) {
+            Profile updatedProfile = profileService.save(event.getProfile());
+            refreshPersistence();
+
+            int value = ((Map<String, Integer>) updatedProfile.getProperty("pageView")).get("acme");
+            Assert.assertEquals(1, value, 0.0);
+        } else {
+            throw new IllegalStateException("Profile was not updated");
```
jkevan, line 89: We should use Assert.fail("the message") instead of Exceptions when something goes wrong in the test. to be fixed on the other test cases too.

itests/src/test/java/org/apache/unomi/itests/IncrementPropertyIT.java
```diff
@@ -0,0 +1,411 @@
...
+        profileService.delete(profile.getItemId(), false);
+    }
+
+    @Test
+    public void testIncrementNotExistingProperty() throws InterruptedException {
+        int eventCode = buildAction("pageView.acme", null, null, null);
+
+        if (eventCode == EventService.PROFILE_UPDATED) {
+            Profile updatedProfile = profileService.save(event.getProfile());
+            refreshPersistence();
```
jkevan, line 84: This is not necessary, you dont need to store the profile in test, just check that the profile from the event is correct.
The save is done automatically by the ContextServlet and it is not required to insure the profile contains the correct values.
You can remove this code here and also in the other test cases.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/IncrementPropertyAction.java
jkevan: I would suggest to test that the nestedProperty is not null and of type Integer.

jkevan (changes requested):

jkevan (approved):

PR #268, opened 2021-03-24
UNOMI-448 : Create generic action to copy properties to profile

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-448


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        // copy props from the specified level (default is: target.properties)
+        try {
+            Object targetProperties = BeanUtilsBean.getInstance().getPropertyUtils().getProperty(event, rootProperty);
+            if (targetProperties instanceof Map) {
+                propsToCopy.putAll((Map) targetProperties);
+            }
+        } catch (Exception e) {
+            // Ignore
```
jkevan, line 104: could be cool to log something like: "Unable to extract properties to be copied from the event to the profile using root property: {}"

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public int execute(Action action, Event event) {
+        boolean changed = false;
+
+        for (Map.Entry<String, Object> entry : getEventPropsToCopy(action, event).entrySet()) {
+            // propType Check
+            PropertyType propertyType = profileService.getPropertyType(entry.getKey());
+            String propertyName = "properties." + entry.getKey();
+            if (propertyType != null && propertyType.isMultivalued()) {
+                changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValue");
```
jkevan, line 54: Doesnt seem's right, in case the event contains an array this will add the array as a value to the existing property. may should use "addValues" instead of "addValue"

Input:
- event input prop: toto = ["newValue"]
- propType: multiple
- profile.properties.toto = ["existingValue"]

Output: 
- profile.properties.toto = ["existingValue", ["newValue"]]

Should be:
- profile.properties.toto = ["existingValue", "newValue"]

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+            } else {
+                Object profileProperty = event.getProfile().getProperty(entry.getKey());
+                if (profileProperty == null) {
+                    if (entry.getValue() instanceof List) {
+                        changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValues");
+                    } else {
+                        changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "alwaysSet");
+                    }
```
jkevan, line 62: This Should depends on the propType, in case the propType exist but is not multivalued you have an issue here:

Input:
- event input prop: toto = ["newValue"]
- propType: singleValue
- profile.properties.toto = null

Output:
-profile.properties.toto = ["newValue"]

Should be:
- FAIL: cannot copy a multiValue from event to a propType that is not multi valued

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public int execute(Action action, Event event) {
+        boolean changed = false;
+
+        for (Map.Entry<String, Object> entry : getEventPropsToCopy(action, event).entrySet()) {
+            // propType Check
+            PropertyType propertyType = profileService.getPropertyType(entry.getKey());
+            String propertyName = "properties." + entry.getKey();
+            if (propertyType != null && propertyType.isMultivalued()) {
```
jkevan, line 53: I would suggest to handle also the case where propertyType != null && !propertyPe.isMultivalued().
because we also need some control here (check next comments that explain why)

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+                    if (entry.getValue() instanceof List) {
+                        changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValues");
+                    } else {
+                        changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValue");
+                    }
+                } else {
+                    if (entry.getValue() instanceof List) {
+                        logger.warn("A single property named {} is already set on the profile.", entry.getKey());
+                    } else {
+                        changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "alwaysSet");
```
jkevan, line 73: We should consider the strategy from action config here.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public int execute(Action action, Event event) {
+        boolean changed = false;
+
+        for (Map.Entry<String, Object> entry : getEventPropsToCopy(action, event).entrySet()) {
+            // propType Check
+            PropertyType propertyType = profileService.getPropertyType(entry.getKey());
+            String propertyName = "properties." + entry.getKey();
```
jkevan, line 52: Here we should check if propertyType have the mandatorySystemTag from action config, if the mandatorySystemTag is provided, then it's also mean that that a propertyType is mandatory for doing the copy.

jkevan (changes requested):

itests/src/test/java/org/apache/unomi/itests/CopyPropertiesActionIT.java
```diff
@@ -0,0 +1,312 @@
...
+
+        event.setProperty("urlParameters", properties);
+
+        int changes = eventService.send(event);
+
+        LOGGER.info("Changes of the event : {}", changes);
+
+        Assert.assertTrue(changes == 0);
+    }
+}
```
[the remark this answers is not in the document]
jkevan: Still not done

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -0,0 +1,116 @@
...
+                if (propertyType.isMultivalued()) {
+                    changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValues");
+                } else if (entry.getValue() instanceof List) {
+                    logger.error("The property {} should contains a single value as declared in the property types", entry.getKey());
+                } else {
+                    changed =
+                            changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), singleValueStrategy);
+                }
+            }
+        }
```
jkevan, line 84: can be simplified: (also we should avoid logging info from the event to avoid creating log injection backdoor)

```
else {
     boolean multipleIsExpected = previousValue != null ? previousValue instanceof List : propertyType.isMultivalued();

      if (multipleIsExpected) {
            changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), "addValues");
      } else if (entry.getValue() instanceof List) {
            logger.error("Impossible to copy the property of type List to the profile, either a single value already exist on the profile or the property type is declared as a single value property. Enable debug log level for more information");
            if (logger.isDebugEnabled()) {
                  logger.debug("cannot copy property {}, because it's a List", entry.getKey())
            }
      } else {
            changed = changed || PropertyHelper.setProperty(event.getProfile(), propertyName, entry.getValue(), singleValueStrategy);
      }
}
```

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
jkevan: Still required

jkevan (changes requested):
Overall logic and algorithm is now good, still some small improvement can be made.

jkevan (approved):
Some test are creating propertyTypes, we need to remove them after each test to avoid conflicting and not interfer on all next tests that may be run in the suite.
Globally speaking an ITest should always the test data after it's finished

PR #269, opened 2021-03-26
UNOMI-451 : Use bus service rest server

jkevan:
…oad server when it's necessary and avoid duplicated features in the bus


dgriffon (approved):

sergehuber (approved):
Thanks Kevan for fixing this. Could you however attach this to an existing ticket or create a new one?

jkevan:
> Thanks Kevan for fixing this. Could you however attach this to an existing ticket or create a new one?

It's already created, and current PR ticket number is the one I created for this occasion.

PR #274, opened 2021-03-31
UNOMI-435 Migrate existing events scope to source

anatol-sialitski:
This PR contains the migration from v1.5.0 to v2.0.0 which updates mapping for an index with prefix `context-event`. Adds the `sourceId` field and copies value from the `scope` field to it.

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private void updateMapping(final String indexName, final CloseableHttpClient httpClient) throws IOException {
+        HttpPut httpPut = new HttpPut(esAddress + "/" + indexName + "/_mapping");
+
+        httpPut.addHeader("Accept", "application/json");
+        httpPut.addHeader("Content-Type", "application/json");
+
+        String request = "{\n" +
+                "\"properties\": {\n" +
+                " \"sourceId\": {\n" +
```
sergehuber, line 94: Seems like the analyzer is missing here.

sergehuber (changes requested):
Hello @anatol-sialitski thanks for the PR. I found a small issue should be easy to fix.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
anatol-sialitski: fixed

sergehuber (dismissed):
Thanks for adding the analyzer so quickly ! I'll merge it.

sergehuber:
Oops I spoke too fast it seems there is a problem with the integration tests, could you check it please ?

anatol-sialitski:
@sergehuber 
IT tests failed for Java 11, but work for Java 8. Looks like delay is too short to process an event and to refresh index during tests running.

sergehuber (approved):
Ok looks good to me will merge and test again

PR #280, opened 2021-04-04
UNOMI-457: Expose refresh policy

liatiusim:
Jira ticket - https://issues.apache.org/jira/browse/UNOMI-457
Add the ability to control the refresh policy when saving an item


sergehuber:
Hi @liatiusim thanks for the PR. Could you explain what the code in the SetEventOccurenceAction is used for ? Taken out of context it is not very clear.

sergehuber (approved):
Ok I understand now that the code in the action is not needed since the query is up to date when event is consistent. PR looks good to me I'll merge it.

PR #281, opened 2021-04-04
UNOMI-456: Add option to query with exclude fields

liatiusim:
Jira ticket: https://issues.apache.org/jira/browse/UNOMI-456
Added the ability to configure properties that will not return in queries. Useful for properties that are for internal use.


sergehuber:
Hi @liatiusim it seems that after merging previous PRs some conflicts have appeared on this PR. Could you fix them please?

sergehuber (commented):
I've finished the review the code looks good to me. Once the conflicts are resolved we can merge it.

PR #287, opened 2021-04-12
handle parameters validation

jsinovassin:

### DIFF
--- a/rest/src/main/java/org/apache/unomi/rest/EventsCollectorEndpoint.java
+++ b/rest/src/main/java/org/apache/unomi/rest/EventsCollectorEndpoint.java
@@ -83,7 +83,7 @@ public class EventsCollectorEndpoint {
 
     @POST
     @Path("/eventcollector")
-    public EventCollectorResponse collectAsPost(EventsCollectorRequest eventsCollectorRequest,
+    public EventCollectorResponse collectAsPost(@NotNull EventsCollectorRequest eventsCollectorRequest,
             @QueryParam("timestamp") Long timestampAsLong) {
         return doEvent(eventsCollectorRequest, timestampAsLong);
     }
@@ -92,10 +92,6 @@ public class EventsCollectorEndpoint {
         Date timestamp = new Date();
         if (timestampAsLong != null) {
             timestamp = new Date(timestampAsLong);
-        }
-
-        if (eventsCollectorRequest == null || eventsCollectorRequest.getEvents() == null) {
-            throw new BadRequestException("No events found");
         }
 
         String sessionId = eventsCollectorRequest.getSessionId();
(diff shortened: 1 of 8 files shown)


rest/src/main/java/org/apache/unomi/rest/EventsCollectorEndpoint.java
```diff
@@ -75,13 +76,15 @@ public Response options() {
...
-    public EventCollectorResponse collectAsGet(EventsCollectorRequest eventsCollectorRequest, @QueryParam("timestamp") Long timestampAsString) {
+    public EventCollectorResponse collectAsGet(@NotNull EventsCollectorRequest eventsCollectorRequest,
+            @QueryParam("timestamp") Long timestampAsString) {
         return doEvent(eventsCollectorRequest, timestampAsString);
     }
 
     @POST
     @Path("/eventcollector")
-    public EventCollectorResponse collectAsPost(EventsCollectorRequest eventsCollectorRequest, @QueryParam("timestamp") Long timestampAsLong) {
+    public EventCollectorResponse collectAsPost(EventsCollectorRequest eventsCollectorRequest,
```
jkevan, line 86: Should be annotate also with NotNull for Post

rest/src/main/java/org/apache/unomi/rest/validation/JAXRSBeanValidationInInterceptorOverride.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+/**
+ * This class allows to replace the class loader use by
+ * javax.el.FactoryFinder.newInstance(FactoryFinder.java:53)
+ * to allow to retrieve com.sun.el.ExpressionFactoryImpl which is present in the class loader of this module
+ */
+public class JAXRSBeanValidationInInterceptorOverride extends JAXRSBeanValidationInInterceptor {
```
jkevan, line 30: It's not used :D
We should use it in the RestServer instead of the JAXRSBeanValidationInInterceptor

jkevan (changes requested):
Seem's good, may be we should target the unomi master instead of 1.5.x ?

rest/src/main/java/org/apache/unomi/rest/EventsCollectorEndpoint.java
dgriffon: Also, if `eventsCollectorRequest` is set as `Not null` statement in [if](https://github.com/apache/unomi/pull/287/files#diff-6431c225a095ce6c2510037b49f2334e82807587b0d381f74cfd42a6e094aea6R97) can be removed

jkevan: Will do in separate PR

jkevan (approved):

PR #288, opened 2021-04-14
UNOMI-463 : add annotations to limit the autorised characters

jsinovassin:

### DIFF
--- a/api/src/main/java/org/apache/unomi/api/utils/ValidationPattern.java
+++ b/api/src/main/java/org/apache/unomi/api/utils/ValidationPattern.java
@@ -18,5 +18,5 @@ package org.apache.unomi.api.utils;
 
 public class ValidationPattern {
 
-    public static final String PATTERN = "^(\\w|[-_@]){0,50}$";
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/apache/unomi/api/utils/ValidationPattern.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.api.utils;
+
+public class ValidationPattern {
+
+    public static final String PATTERN = "^(\\w|[-_@]){0,50}$";
```
jkevan, line 21: We should use a more meaningful name for this pattern, like TEXT_VALID_CHARACTERS_PATTERN or ID_VALIDATION_PATTERN.

jkevan (changes requested):

jkevan (approved):

dgriffon (approved):

PR #292, opened 2021-04-16
UNOMI-463 : add exception mapping to send error 400

jsinovassin:


jkevan (changes requested):
Could be great to use latest integration test and add a check to the responses to see that we correctly have the expected message. Currently we only test the status code, but status code 400 could be triggered by something else than the validation, so checking the message would be great.

jkevan:
> Could be great to use latest integration test and add a check to the responses to see that we correctly have the expected message. Currently we only test the status code, but status code 400 could be triggered by something else than the validation, so checking the message would be great.

We will do it in a separate PR

PR #293, opened 2021-04-16
UNOMI-463: Add integrations test on top of input validations

jkevan:


sergehuber (commented):
Thanks for the PR, looks good to me but we will have to wait on UNOMI-467 to be fixed before we can merge it.

PR #294, opened 2021-04-16
UNOMI-463 :add minimale log for validation exception

jsinovassin:

### DIFF
--- a/rest/src/main/java/org/apache/unomi/rest/exception/ValidationExceptionMapper.java
+++ b/rest/src/main/java/org/apache/unomi/rest/exception/ValidationExceptionMapper.java
@@ -29,11 +29,14 @@ import javax.ws.rs.ext.Provider;
 @Provider
 @Component(service = ExceptionMapper.class)
 public class ValidationExceptionMapper implements ExceptionMapper<ConstraintViolationException> {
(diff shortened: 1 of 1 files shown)


rest/src/main/java/org/apache/unomi/rest/exception/ValidationExceptionMapper.java
```diff
@@ -29,11 +29,14 @@
 @Provider
 @Component(service = ExceptionMapper.class)
 public class ValidationExceptionMapper implements ExceptionMapper<ConstraintViolationException> {
+    private static final Logger logger = LoggerFactory.getLogger(ValidationExceptionMapper.class.getName());
 
     @Override
     public Response toResponse(ConstraintViolationException exception) {
-        return Response.status(Response.Status.BAD_REQUEST).header("Content-Type", MediaType.TEXT_PLAIN)
-                .entity("Invalid received data").build();
+        exception.getConstraintViolations().forEach(constraintViolation -> logger
```
jkevan, line 36: we should propose to enable debug log level in the error message.
Then we should display the full error message in debug log, with the input value.

jkevan (changes requested):

jkevan (approved):

PR #296, opened 2021-04-19
UNOMI-468 : consider mapping in copy properties action

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-468


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/CopyPropertiesAction.java
```diff
@@ -112,4 +113,9 @@ public int execute(Action action, Event event) {
 
         return propsToCopy;
     }
+
+    private String getPropertyName(String propertyName) {
```
sergehuber, line 117: I wonder if we should rename this to resolvePropertyName, getPropertyName seems weird to me.

sergehuber (approved):
Hi Jonathan, thanks for the PR! 

I just put a small comment, it seems like one of the method's name is a bit ambiguous

PR #297, opened 2021-04-20
request size check

jkevan:


dgriffon (approved):

rest/src/main/java/org/apache/unomi/rest/service/impl/RestServiceUtilsImpl.java
```diff
@@ -64,7 +64,7 @@ public String getProfileIdCookieValue(HttpServletRequest httpServletRequest) {
         if (cookies != null) {
             for (Cookie cookie : cookies) {
                 if (configSharingService.getProperty("profileIdCookieName").equals(cookie.getName())) {
-                    localBeanValidationProvider.get().validateBean(new CookieWrapper(cookie.getValue()));
```
jsinovassin, line 67: We can rename the variable to beanValidationService

jsinovassin (approved):

sergehuber:
Hello Kevan, 

Is there a ticket for this PR? Maybe attach it to one of the existing ones ?

rest/src/main/java/org/apache/unomi/rest/exception/InvalidRequestExceptionMapper.java
```diff
@@ -0,0 +1,41 @@
...
+@Component(service = ExceptionMapper.class)
+public class InvalidRequestExceptionMapper implements ExceptionMapper<InvalidRequestException> {
+    private static final Logger logger = LoggerFactory.getLogger(InvalidRequestExceptionMapper.class.getName());
+
+    @Override
+    public Response toResponse(InvalidRequestException exception) {
+        logger.error(exception.getMessage(), exception);
+
+        return Response.status(Response.Status.BAD_REQUEST).header("Content-Type", MediaType.TEXT_PLAIN)
+                .entity("Request rejected by the server because: " + exception.getResponseMessage()).build();
```
sergehuber, line 39: I wonder if we should send the exception message because it might contain some sensitive data no?

sergehuber (commented):
Hi Kevan thanks for the PR. Looks good to me apart from the Exception mapper that probably disclosed a bit too much information (see inline comment)

Edit: my bad I misread the code, please disregard.

rest/src/main/java/org/apache/unomi/rest/exception/InvalidRequestExceptionMapper.java
sergehuber: Maybe we could generate a debug log with more detailed information ?

sergehuber: Forget what I said I was reading the code wrong about the message.

rest/src/main/java/org/apache/unomi/rest/service/impl/RestServiceUtilsImpl.java
jkevan: Thx I will rename it

PR #300, opened 2021-04-21
[UNOMI-469] + [UNOMI-470] Actions that add, remove value to profile property, ncrementing, Decrementing profile property

atompie:
PR Title format: 

    [UNOMI-469] Actions that add value to profile property

Action that add and remove value from property list. Existing actions do not remove values. Adding is also throwing errors when there is a type conflict that means property list consists integers and we add strings.

  [UNOMI-470] Incrementing, Decrementing profile property

Action to Increment, Decrement profile property. Currently there is not default action. You could probably do it by white listing MVEL expressions.

With this plugin we want to deliver a set of basic actions to use within TRACARDI User Interface for Unomi.

Info about Tracardi
https://www.facebook.com/TRACARDI/

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -39,6 +39,7 @@
         <module>past-event</module>
         <module>tracked-event</module>
         <module>kafka-injector</module>
+        <module>tracardi-plugins</module>
     </modules>
 
     <dependencies>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -41,8 +41,10 @@
         <dependency>
             <groupId>javax.validation</groupId>
             <artifactId>validation-api</artifactId>
-            <version>1.1.0.Final</version>
-            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-validator</artifactId>
         </dependency>
     </dependencies>
 
--- a/api/src/main/java/org/apache/unomi/api/ContextRequest.java
+++ b/api/src/main/java/org/apache/unomi/api/ContextRequest.java
@@ -24,7 +24,9 @@ import org.apache.unomi.api.services.EventListenerService;
 import org.apache.unomi.api.services.PersonalizationService;
 import org.apache.unomi.api.services.ProfileService;
 import org.apache.unomi.api.services.RulesService;
+import org.apache.unomi.api.utils.ValidationPattern;
 
+import javax.validation.constraints.Pattern;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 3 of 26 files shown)


plugins/pom.xml
```diff
@@ -39,6 +39,7 @@
         <module>past-event</module>
         <module>tracked-event</module>
         <module>kafka-injector</module>
+        <module>tracardi-plugins</module>
```
sergehuber, line 42: I would prefer we do not add a "client" specific plugin here. Rather these actions, if we keep them (see below), should become part of the base plugin.

sergehuber (changes requested):
Hello @atompie thank you for your PR. A few things I think that should be modified. 

1. Please make the PR against the master, we always do it like this and then cherry-pick to the target branches as need be.
2. Why did you add new actions rather than improving the existing ones (UpdatePropertiesAction and SetPropertiesAction) ? 
3. I put a comment in the code about this but I don't think this should be part of a client-specific plugin but rather included in the base plugin as I think it should be core functionality.
4. To make sure there is no breakage I think it would be best to include integration tests for this new functionality and to make sure we don't break existing functionality.

Just to clarify, UpdatePropertiesAction is the action targeted to clients, while SetPropertiesAction is more designed for internal use (UpdatePropertiesAction actually uses the SetPropertiesAction). This is by design because UpdatePropertiesAction is associated to an "updateProperties" event type that is protected.

Please don't hesitate to come back to me. 

Regards,
  Serge...

sergehuber:
Also I think it would be interesting to sort of regroup the IncrementPropertyAction with some of this.

jkevan (changes requested):
Hello thank you for the proposition it's clearly interesting.

As Serge mentioned there is already the SetPropertyAction that cover a lot of cases.
- It is already capable of adding values to list using the "setPropertyStrategy" param (addValue, addValues)
So it could be interesting to add new strategies instead of new actions.
- new strategies like: removeValue, removeValues could be implemented in the PropertyHelper.setProperty.
That way we update the existing mechanism.

For the incrementation, we recently develop a new action: IncrementPropertyAction
It's capable of incrementing any properties with values coming from the event, but also capable of simply incrementing +1 any props.
It could be interesting to reuse this one in your project.

The only missing part in actual Unomi would the the decrement action, I dont think we have that currently.
But we could improve the existing IncrementPropertyAction to support decrement.

Do you think it would be possible to modify the current PR to do this ?
It could be an advantage for you, as it is already existing mechanisms that are tested on our sides, and we do maintain them.
Overall I think there is not much to do to support what you need with existing system, and I more in favor of improving the existing systems than create similar ones in the same project as it could confuse peoples at the end.

Regards,
Kevan.

atompie:
Unfortunately I am lacking time recently working on the GUI. So can't work on this at this time. If you would like to continue with changes to this PR please take over :)

jkevan:
Ok no problem,
Are you ok if we close this pull request ?
In the future we will probably implement this features like we describe, but it's currently not our priority.

atompie:
Sure.

sergehuber:
Closing as approved.

PR #301, opened 2021-04-26
UNOMI-472-fix conditionEvaluator inconsistent with elastic for condition when p…

giladw:
…roperty  is not exist and op is "notIn"

PR Title format: 

    [UNOMI-XXX] Pull request title with JIRA reference

**Please** add a meaningful description for your change here

----
when using PropertyCondition, i noticed inconsistency when :
1. creating a segment + sending an event the would enter the profile to a segment
2. creating a profile that matches condition + creating the segment.

for example that reproduces :
{
       "type": "profilePropertyCondition",
       "parameterValues":

{          "propertyName": "properties.productIdsList",          "comparisonOperator": "notIn",          "propertyValuesInteger": [92713211]        }

}

if a profile dosnet have the field properties.productIdsList it would be as described

**Please** following this checklist to help us incorporate your contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/UNOMI) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
 - [x] Format the pull request title like `[UNOMI-XXX] - Title of the pull request`
 - [x] Provide integration tests for your changes, especially if you are changing the behavior of existing code or adding
       significant new parts of code.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why. 
       Copy the description to the related JIRA issue
 - [x] Run `mvn clean install -P integration-tests` to make sure basic checks pass. A more thorough check will be 
        performed on your pull request automatically.
 
Trivial changes like typos do not require a JIRA issue (javadoc, project build changes, small doc changes, comments...). 
 
If this is your first contribution, you have to read the [Contribution Guidelines](https://unomi.apache.org/contribute.html)

If your pull request is about ~20 lines of code you don't need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) 
if you are unsure please ask on the developers list.

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

### DIFF
--- a/plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
+++ b/plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
@@ -206,7 +206,7 @@ public class PropertyConditionEvaluator implements ConditionEvaluator {
         if (op == null) {
             return false;
         } else if (actualValue == null) {
-            return op.equals("missing");
+            return op.equals("missing")|| op.equals("notIn") || op.equals("notEquals");
         } else if (op.equals("exists")) {
             if (actualValue instanceof List) {
                 return ((List) actualValue).size() > 0;
--- a/itests/src/test/java/org/apache/unomi/itests/ConditionEvaluatorIT.java
+++ b/itests/src/test/java/org/apache/unomi/itests/ConditionEvaluatorIT.java
@@ -140,6 +140,9 @@ public class ConditionEvaluatorIT extends BaseIT {
         assertTrue(eval(builder.profileProperty("properties.age").in(31, 30).build()));
         assertTrue(eval(builder.profileProperty("properties.age").notIn(25, 26).build()));
         assertFalse(eval(builder.profileProperty("properties.age").notIn(25, 30).build()));
+
+        assertTrue(eval(builder.profileProperty("properties.fieldNotExists").notIn(25, 30).build()));
+        assertTrue(eval(builder.profileProperty("properties.fieldNotExists").notEqualTo(1).build()));
     }
 
     @Test


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
```diff
@@ -206,7 +206,7 @@ protected boolean isMatch(String op, Object actualValue, String expectedValue, O
         if (op == null) {
             return false;
         } else if (actualValue == null) {
-            return op.equals("missing");
+            return op.equals("missing")|| op.equals("notIn") || op.equals("notEquals");
```
jkevan, line 209: What happen if:

expectedValue: null
actualValue: null
op: notEquals

It's look like it will return true, but it should return false in that case.

jkevan (changes requested):
Hello, thx for the pull request.

It could have been solved by using a more complete condition:
properties.productIdsList isMissing OR properties.productIdsList notIn [92713211]

But overall ok with this changes, appart from the notEquals handling that could lead to inconsistent results (see comment in the PR code)

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
giladw: Thanks @jkevan 
i dont think we can have both  expectedValue=null && op=notEquals
like that:
{
        "type": "profilePropertyCondition",
        "parameterValues": {
          "propertyName": "properties.countryCode",
          "comparisonOperator": "notEquals",
          "propertyValue": "null"
        }
}

in such case we would use comparisonOperator=missing
{
        "type": "profilePropertyCondition",
        "parameterValues": {
          "propertyName": "properties.countryCode",
          "comparisonOperator": "missing"
        }
}

jkevan:
Thank you for the answers.
If we are sure that elasticsearch have the same behavior when evaluating this condition I would say we are good.
But I'm not sure, do you know if ElasticSearch queries are also handling this case this way ?

sergehuber:
@giladw would love your answer on this or we'll have to close the PR.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
giladw: sorry for the delay, from what i tested, then yes in elastic it handles it this way.
the way i test it is have 2 tests:
1. create a segment condition,  then send an event that matches, after validates that the profile is in the segment.
2. create the event first and then create the segment

jkevan: Thx for the clarification

jkevan: But I also found this case that should be handle the same: 'hasNoneOf'

```suggestion
            return op.equals("missing")|| op.equals("notIn") || op.equals("notEquals") || op.equals("hasNoneOf");
```

jkevan (changes requested):

jkevan:
Hello @giladw , could you consider the proposition to handle "hasNoneOf" ? so we can merge this PR and move one.

Thanks you,
Best regards.

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/PropertyConditionEvaluator.java
giladw: Hi, sorry im working on a different project at the moment and i will not have time to push new commits

Fgerthoffert:
Hi @giladw,

As per our discussion on Slack, we're going to merge your PR as-is and do the requested modification in a follow-up PR.

Thanks,

cc: @sergehuber

PR #313, opened 2021-06-23
UNOMI-474 Add priorities to GraphQL field visibility providers

pmi:
- Added priority field to graphql visibility provider interface
- Created composite graphql visibility provider to combine multiple visibility providers for graphql schema generation
- Added 2 visibility providers to graphql-providers-sample feature as example

### DIFF
--- a/samples/graphql-providers/src/main/java/org/apache/unomi/graphql/providers/sample/CDPVisibilityOnlyProvider.java
+++ b/samples/graphql-providers/src/main/java/org/apache/unomi/graphql/providers/sample/CDPVisibilityOnlyProvider.java
@@ -25,6 +25,12 @@ import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
 
+/**
+ * Example graphql visibility provider
+ * It will block:
+ * - CDP_SegmentInput's view field as well as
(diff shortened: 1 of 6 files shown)


samples/graphql-providers/src/main/java/org/apache/unomi/graphql/providers/sample/CDPVisibilityOnlyProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import graphql.schema.visibility.BlockedFields;
+import graphql.schema.visibility.GraphqlFieldVisibility;
+import org.apache.unomi.graphql.providers.GraphQLFieldVisibilityProvider;
+import org.apache.unomi.graphql.providers.GraphQLProvider;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+
+@Component(immediate = true, service = GraphQLProvider.class)
```
sergehuber, line 28: Could you please add a class description for this provider ? Thanks a lot

sergehuber (changes requested):
Hi @pmi thanks a lot for the PR. If you could just add a description to the sample provider that would be great! Otherwise it looks like it's ready to go.

PR #326, opened 2021-07-23
DMF-4797 : separate last visit logic from set property action

jsinovassin:
https://jira.jahia.org/browse/DMF-4797


jkevan:
This have been implemented in: https://github.com/apache/unomi/pull/329
and lastVisit calcul have been reworked.

PR #329, opened 2021-07-30
UNOMI-338, UNOMI-482: better implementation for lastVisit, firstVisit…

jkevan:
… and previousVisit properties

I would like to propose this new Implem for the handling of the lastVisit, firstVisit and previousVisit properties on the profile.

The precedent implem was not logical and error prone.
previous algorithm:

profile:
- firstVisit: 11/05/2021
- previousVisit: 13/05/2021
- lastVisit: 16/05/2021
In case Event comes with a timeStamp: 10/05/2021, but the event is submitted to the system at 20/05/2021.
Then the result was:
- firstVisit: 11/05/2021
- previousVisit: 16/05/2021
- lastVisit: 20/05/2021

This seem's not logical because the eventTimestamp is only source of truth regarding the chronology of the events and the visits. So we should not use the current date that the event is processed into the system.
So I propose this result instead:
- firstVisit: 10/05/2021
- previousVisit: 13/05/2021
- lastVisit: 16/05/2021

My implem is updating the 3 dates to respect the chronology of the events received (previous implem was only updating lastVisit with no sense values).
So depending on the event timeStamp, one of multiples of this 3 props will be updated accordingly to the eventTimeStamp.
The profile will hold this informations updated.

It have multiple benefits:
- more logic and easy to understand (only one source of truth for the chronology of events)
- better handling of events coming from the past (using pasted dates will be more easy to handle by the system)
- segmentation will work better for segment with condition on this 3 props.

Also this PR contains a simple improvement already asked for: https://issues.apache.org/jira/browse/UNOMI-482
To be able to set the currentDate in the SetPropertyAction.
SetPropertyAction with setPropertyValue: 'now' is now deprecated, I propose to use 'setPropertyValueCurrentEventTimestamp' or 'setPropertyValueCurrentDate' instead of 'setPropertyValue' that is more suitable for this need.

Also this PR contains the integration tests necessary to tests all this changes.

### DIFF
--- a/itests/src/test/java/org/apache/unomi/itests/BasicIT.java
+++ b/itests/src/test/java/org/apache/unomi/itests/BasicIT.java
@@ -50,6 +50,7 @@ import org.slf4j.LoggerFactory;
 import javax.inject.Inject;
 import java.io.File;
 import java.io.IOException;
+import java.text.SimpleDateFormat;
 import java.util.*;
 
 
@@ -165,6 +166,8 @@ public class BasicIT extends BaseIT {
                 ContentType.create("application/json")));
         TestUtils.RequestResponse requestResponsePageView1 = executeContextJSONRequest(requestPageView1, SESSION_ID_3);
         String profileIdVisitor1 = requestResponsePageView1.getContextResponse().getProfileId();
+        String lastVisit = (String) requestResponsePageView1.getContextResponse().getProfileProperties().get("lastVisit");
+        Assert.assertNotNull("Context profile properties should contains a lastVisit property", lastVisit);
         Thread.sleep(1000);
 
         // Initialize VISITOR_1 properties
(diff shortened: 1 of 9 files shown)


itests/src/test/java/org/apache/unomi/itests/BasicIT.java
```diff
@@ -265,7 +278,7 @@ private ContextRequest getContextRequestWithLoginEvent(CustomItem sourceSite, Ma
         contextRequest.setSource(sourceSite);
         contextRequest.setRequireSegments(false);
         contextRequest.setEvents(Collections.singletonList(loginEvent));
-        contextRequest.setRequiredProfileProperties(Arrays.asList(FIRST_NAME, LAST_NAME, EMAIL));
+        contextRequest.setRequiredProfileProperties(Arrays.asList(FIRST_NAME, LAST_NAME, EMAIL, "firstVisit", "lastVisit", "previousVisit"));
```
sergehuber, line 281: Could we use constants for those property names too ?

itests/src/test/java/org/apache/unomi/itests/BasicIT.java
```diff
@@ -290,7 +303,7 @@ private ContextRequest getContextRequestWithPageViewEvent(CustomItem sourceSite,
         contextRequest.setSource(customPageItem);
         contextRequest.setRequireSegments(false);
         contextRequest.setEvents(Collections.singletonList(pageViewEvent));
-        contextRequest.setRequiredProfileProperties(Arrays.asList(FIRST_NAME, LAST_NAME, EMAIL));
+        contextRequest.setRequiredProfileProperties(Arrays.asList(FIRST_NAME, LAST_NAME, EMAIL, "firstVisit", "lastVisit", "previousVisit"));
```
sergehuber, line 306: Some comments as before, maybe use constants here ?

sergehuber (changes requested):
Hi Kevan, 
Thanks a lot for the PR. Looks mostly good to me, just had a small request about using constants in the tests to make it consistent.

I wonder if we shouldn't use constants everywhere actually but I don't want to request too many changes :) 

Regards,
  Serge...

sergehuber (approved):
Looks good to me now, thanks for the changes !

PR #330, opened 2021-08-10
UNOMI-502: do not validate disabled segments/scoring to be able to sa…

jkevan:
…ve incomplete objects (only in case they are disabled)


jsinovassin (approved):

sergehuber (approved):
Thanks for the PR @jkevan . Looks good to me.

PR #331, opened 2021-08-12
UNOMI-503: disable cluster sync because unomi bundles states in clust…

jkevan:
…er are not correct


sergehuber (approved):
Thanks @jkevan for the PR. Although it would be better to replace this with proper feature support in the future I think this solution is probably the safest at this point.

PR #336, opened 2021-09-08
UNOMI-506 Delete aliases when a profile is deleted

anatol-sialitski:
Delete aliases when a profile is deleted


sergehuber (approved):
Thanks for the PR Anatol it looks good to me I'll merge it.

PR #337, opened 2021-09-09
UNOMI-508 provide Unomi Groovy actions endpoint

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-508


extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/annotations/Action.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.groovy.actions.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Action {
```
sergehuber, line 23: Could you please add some basic Javadoc comments for the class and the parameters of the annotation so that they might show up in IDEs ?

extensions/groovy-actions/karaf-kar/pom.xml
```diff
@@ -61,7 +69,11 @@
             <artifactId>jsoup</artifactId>
             <version>1.13.1</version>
         </dependency>
-
+        <dependency>
+            <groupId>com.sun.activation</groupId>
```
sergehuber, line 73: Why is this needed ?

extensions/groovy-actions/rest/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-transports-http</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-rs-security-cors</artifactId>
```
sergehuber, line 71: Shouldn't this be with <scope>provided</scope> ?

extensions/groovy-actions/rest/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
```
sergehuber, line 75: Same here : provided ?

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/GroovyAction.java
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.groovy.actions;
+
+import org.apache.unomi.api.Metadata;
+import org.apache.unomi.api.MetadataItem;
+
+/**
+ * Object which represents a grrovy action
```
sergehuber, line 23:
```suggestion
 * Object which represents a Groovy action in persistence service
```

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/GroovyActionDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+    public void setBundleContext(BundleContext bundleContext) {
+        this.bundleContext = bundleContext;
+    }
+
+    public String getPrefix() {
+        return GROOVY_PREFIX;
+    }
+
+    public Integer execute(Action action, Event event, String actionName) {
+        GroovyAction groovyScript = persistenceService.load(actionName, GroovyAction.class);
```
sergehuber, line 63: Ideally we should have some kind of cache for the scripts, possibly in the GroovyActionService. 

So we would instead do something like groovyActionService.getScript(actionName), and in the service manage the cache and reload from the service with a thread on a regular basis (since other cluster nodes could deploy groovy scripts), as is done in other service implementations.

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/annotations/Parameter.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.groovy.actions.annotations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Parameter {
```
sergehuber, line 23: Same thing here, please add Javadoc to explain how to use this

extensions/groovy-actions/rest/src/main/java/org/apache/unomi/groovy/actions/rest/GroovyActionsEndPoint.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    @POST
+    @Path("/")
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response save(@Multipart(value = "file") Attachment file) {
+        try {
+            String filename = file.getContentDisposition().getParameter("filename");
+
+            java.nio.file.Path path = Paths.get("/tmp/" + filename);
```
sergehuber, line 80: I don't think this way of retrieving the temporary directory is correct. Instead we should use something like this: https://attacomsian.com/blog/java-create-temporary-directory#:~:text=In%20Java%207%20and%20higher%2C%20you%20can%20use,and%20print%20its%20absolute%20path%20on%20the%20console. and create a "unomi-groovy-actions" directory to make sure we don't mess up the main temp directory.

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/services/GroovyActionsService.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.groovy.actions.services;
+
+import java.io.File;
+
+/**
+ * Short description of the interface
```
sergehuber, line 22: It is the convention that all interfaces have minimal Javadoc documentation in Unomi. Please add some documentation for the class and methods. Thank you

extensions/groovy-actions/services/src/main/resources/META-INF/cxs/mappings/groovyAction.json
```diff
@@ -0,0 +1,20 @@
...
+          "fields": {
+            "keyword": {
+              "type": "keyword",
+              "ignore_above": 256
+            }
+          }
+        }
+      }
+    }
+  ]
```
sergehuber, line 19: Since this type will not change, I think it would be best to put the full mapping for it.

sergehuber (changes requested):
Hello Jonathan, thank you very much for this PR. However it is missing a Unomi ticket in the title (please create one if it doesn't exist yet). I also found some issues in the code and it is currently missing integration tests.

I will review it once you submit further changes.

Best regards,
  Serge...

extensions/groovy-actions/karaf-kar/pom.xml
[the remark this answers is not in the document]
jsinovassin: I added it because of this error message: org.apache.felix.resolver.reason.ReasonException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=unomi-groovy-actions; type=karaf.feature; version="[1.6.0.SNAPSHOT,1.6.0.SNAPSHOT]"; filter:="(&(osgi.identity=unomi-groovy-actions)(type=karaf.feature)(version>=1.6.0.SNAPSHOT)(version<=1.6.0.SNAPSHOT))" [caused by: Unable to resolve unomi-groovy-actions/1.6.0.SNAPSHOT: missing requirement [unomi-groovy-actions/1.6.0.SNAPSHOT] osgi.identity; osgi.identity=org.apache.unomi.groovy-actions-rest; type=osgi.bundle; version="[1.6.0.SNAPSHOT,1.6.0.SNAPSHOT]"; resolution:=mandatory [caused by: Unable to resolve org.apache.unomi.groovy-actions-rest/1.6.0.SNAPSHOT: missing requirement [org.apache.unomi.groovy-actions-rest/1.6.0.SNAPSHOT] osgi.wiring.package; filter:="(&(osgi.wiring.package=javax.activation)(version>=1.2.0)(!(version>=2.0.0)))"]]

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/GroovyAction.java
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.unomi.groovy.actions;
+
+import org.apache.unomi.api.Metadata;
+import org.apache.unomi.api.MetadataItem;
+
+/**
+ * Object which represents a Groovy action
```
sergehuber, line 23:
```suggestion
 * Object which represents a Groovy action (including its script)
```
```suggestion
 * Object which represents a Groovy action
```

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/utils/Utils.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Utility class
+ */
+public class Utils {
+
+    public static String convertInputStreamToString(InputStream inputStream) throws IOException {
```
sergehuber, line 30: Please do not use your own implementation, there are a LOT of available ones, especially in the Commons IO library.

itests/src/test/java/org/apache/unomi/itests/GroovyActionsServiceIT.java
```diff
@@ -0,0 +1,108 @@
...
+        refreshPersistence();
+    }
+
+    @After
+    public void cleanUp() throws InterruptedException {
+        refreshPersistence();
+    }
+
+    private String loadGroovyAction(String pathname) throws IOException {
+        return Utils.convertInputStreamToString(new FileInputStream(new File(pathname)));
```
sergehuber, line 63: Please replace with IOUtils.toString(InputStream)

itests/src/test/java/org/apache/unomi/itests/GroovyActionsServiceIT.java
```diff
@@ -0,0 +1,108 @@
...
+
+        Assert.assertNotNull(groovyAction);
+
+        groovyActionsService.remove("MyAction");
+
+        Thread.sleep(2000);
+
+        groovyAction = groovyActionsService.getGroovyAction("scriptGroovyAction");
+
+        Assert.assertNull(groovyAction);
```
sergehuber, line 106: Shouldn't we also remove the action from the definition service ?

sergehuber (changes requested):
Please see below for changes, mostly removing the custom copyInputStreamToString with IOUtils.toString()

extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/GroovyActionDispatcher.java
```diff
@@ -0,0 +1,89 @@
...
+        GroovyAction groovyAction = groovyActionsService.getGroovyAction(actionName);
+        if (groovyAction == null) {
+            logger.warn("Couldn't find a Groovy action with name {}, action will not execute !", actionName);
+        } else {
+            try {
+                return new MetricAdapter<Integer>(metricsService, this.getClass().getName() + ".action.groovy." + actionName) {
+                    @Override
+                    public Integer execute(Object... args) throws Exception {
+                        GroovyBundleResourceConnector bundleResourceConnector = new GroovyBundleResourceConnector(bundleContext);
+                        GroovyScriptEngine engine = new GroovyScriptEngine(bundleResourceConnector,
```
sergehuber, line 75: We should probably look into caching this if it is acceptable in Groovy practices.

sergehuber (changes requested):
Found a potential performance issue.

sergehuber (approved):
After discussion, we will handle performance optimization in a subsequent PR. Thanks for the changes, they can now be merged.

PR #340, opened 2021-09-15
 [UNOMI-462] Fix recalculation of Segments containing past event conditions, also …

jkevan:
…provide ITest

This is a PR to fix the recalculation of past event conditions used in the daily Timer.
Fixed issues:
- correctly recalculate linked segments after event count occurrences have been updated on profiles for a given past event auto generated rule
- correctly reset event count occurrences to 0 on profiles that do not have matching events anymore in the time based past event condition.

Also provided an ITest related to this fixes.


sergehuber (approved):
Thanks for the PR it looks good to me now. Thanks for changing it to the BaseIT.keepTrying method too.

PR #341, opened 2021-09-17
Improve groovy actions performances

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-509


sergehuber (approved):
Thanks for improving this, it looks a lot better. I will merge now.

PR #346, opened 2021-09-29
return the result of the groovy action dispatcher

jsinovassin:


sergehuber (approved):
Thanks a lot Jonathan for fixing that. I'll merge it.

PR #357, opened 2021-11-02
UNOMI-510 : add new shared scheduler …

dgriffon:
add new shared scheduler  with a pool a threads that can be run task in parallel and use it on geonames import

see https://issues.apache.org/jira/browse/UNOMI-510 for details


sergehuber (approved):
Looks good to me, thank you for the contribution. 
Regards,
  Serge...

PR #363, opened 2021-11-18
UNOMI-530 : fix maximumEventCount in past event condition

dgriffon:
https://issues.apache.org/jira/browse/UNOMI-530

Covered by integration test.

### DIFF
--- a/itests/src/test/java/org/apache/unomi/itests/SegmentIT.java
+++ b/itests/src/test/java/org/apache/unomi/itests/SegmentIT.java
@@ -393,7 +393,7 @@ public class SegmentIT extends BaseIT {
         scoringElements.add(scoringElement);
         scoring.setElements(scoringElements);
         segmentService.setScoringDefinition(scoring);
-        Thread.sleep(5000);
+        refreshPersistence();
 
         // Send 2 events that match the scoring plan.
         Event testEvent = new Event("test-event-type", null, profile, null, null, profile, timestampEventInRange);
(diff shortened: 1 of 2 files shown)


itests/src/test/java/org/apache/unomi/itests/SegmentIT.java
```diff
@@ -358,4 +360,100 @@ public void testScoringPastEventRecalculation() throws Exception {
...
+        eventService.send(testEvent);
+        refreshPersistence();
+        // 2nd event
+        testEvent = new Event("test-event-type", null, testEvent.getProfile(), null, null, testEvent.getProfile(), timestampEventInRange);
+        eventService.send(testEvent);
+        refreshPersistence();
+
+        // insure the profile is engaged;
+        try {
+            Assert.assertTrue("Profile should have 2 events in the scoring",  (Long) ((Map) testEvent.getProfile().getSystemProperties().get("pastEvents")).get(pastEventCondition.getParameterValues().get("generatedPropertyKey")) == 2);
```
jkevan, line 409: This is not enough to test that the profile is engaged in the scoring, this is just checking that the auto generated rules are working correctly. But also need to check that the score property is correctly set on the profile.

jkevan (approved):

itests/src/test/java/org/apache/unomi/itests/SegmentIT.java
dgriffon: We have another test that is validating the engaged profiles, I wanted to focus that test to what it is supposed to test.

dgriffon: I updated the test to add the check, thank you for the review

jkevan (approved):

PR #365, opened 2021-11-18
UNOMI-532 

dgriffon:
https://issues.apache.org/jira/browse/UNOMI-532
Add a check at refresh level to invalid rules with missing action or invalid condition. 
Covered by unit tests

### DIFF
--- a/services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
+++ b/services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
@@ -277,11 +277,26 @@ public class RulesServiceImpl implements RulesService, EventListenerService, Syn
 
     private List<Rule> getAllRules() {
         List<Rule> rules = persistenceService.getAllItems(Rule.class, 0, -1, "priority").getList();
+        List<Rule> rulesToDisable = new ArrayList<>();
+        List<Rule> enabledRules = new ArrayList<>();
         for (Rule rule : rules) {
-            ParserHelper.resolveConditionType(definitionsService, rule.getCondition(), "rule " + rule.getItemId());
-            ParserHelper.resolveActionTypes(definitionsService, rule);
+            // Check rule integrity
+            boolean isValid = ParserHelper.resolveConditionType(definitionsService, rule.getCondition(), "rule " + rule.getItemId());
+            isValid = isValid && ParserHelper.resolveActionTypes(definitionsService, rule);
+            // exclude enabled invalid rules
+            if (isValid) {
+                enabledRules.add(rule);
+            } else if (rule.getMetadata().isEnabled()) {
+                rulesToDisable.add(rule);
+            }
         }
-        return rules;
+        // Disable invalid rules and store it.
+        rulesToDisable.forEach(rule -> {
+            logger.warn("Rule {} has been disabled due to invalid condition or actions", rule.getItemId());
+            rule.getMetadata().setEnabled(false);
+            persistenceService.save(rule);
+        });
+        return enabledRules;
     }
 
     private Map<String, Set<Rule>> getRulesByEventType(List<Rule> rules) {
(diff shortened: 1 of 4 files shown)


services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
```diff
@@ -393,6 +408,11 @@ public void setRule(Rule rule) {
                 definitionsService.extractConditionBySystemTag(condition, "eventCondition");
             }
         }
+        List<Action> actions = rule.getActions();
+        if (actions == null || actions.isEmpty()) {
+            logger.warn("rule {} won't be saved as it contains no action", rule.getItemId());
+            return;
+        }
```
jkevan, line 415: We need to check if the rule is disabled here, in case it is disabled we could authorize the persitence even if there is no action.
It's like a draft rule.

Also it could be interesting to force the disable but save anyway in case of no action to avoid user loosing there work in case they are using some sort of UI to build the rule, they may not be aware that the rule is not saved at the end.

Also in case we still want to not save the rule, then we should crash an Exception or a validation error message. Because the setRule didn't do the expected behavior.
I think force disabling the rule and save anyway would be the more easy to do. and more logic considering the rest of the fix

jkevan (changes requested):

services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
```diff
@@ -277,11 +277,26 @@ public void refreshRules() {
...
+            // exclude enabled invalid rules
+            if (isValid) {
+                enabledRules.add(rule);
+            } else if (rule.getMetadata().isEnabled()) {
+                rulesToDisable.add(rule);
+            }
         }
-        return rules;
+        // Disable invalid rules and store it.
+        rulesToDisable.forEach(rule -> {
```
sergehuber, line 294: I wonder if we should put a setting around this ?

services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
sergehuber: I think I would be more in favor of raising an exception in this case.

sergehuber (changes requested):
Hello David, 
Thanks a lot for your PR. I put some comments in the code but I also wanted to say that we should probably improve the integration tests to include the changes you have introduced here. I think we already have some tests related to invalid rules so it would just be a matter of improving the tests.

sergehuber:
Also I would appreciate a more descriptive title for the PR :)

services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
dgriffon: Nice catch, get all rules should return always ALL rules

services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
```diff
@@ -277,10 +278,23 @@ public void refreshRules() {
...
-            ParserHelper.resolveConditionType(definitionsService, rule.getCondition(), "rule " + rule.getItemId());
-            ParserHelper.resolveActionTypes(definitionsService, rule);
+            // Check rule integrity
+            boolean isValid = ParserHelper.resolveConditionType(definitionsService, rule.getCondition(), "rule " + rule.getItemId());
+            isValid = isValid && ParserHelper.resolveActionTypes(definitionsService, rule);
+            // check if rule status has changed
+            if (rule.getMetadata().isEnabled() != isValid) {
+                rule.getMetadata().setEnabled(isValid);
+                rulesToSwitch.add(rule);
+            }
```
jkevan, line 290: Something is wrong here, in case:
- rule enabled: false
- rule valid: true

Then it will force enable the rule and save it as enabled, soometimes users just want to create draft rule that could be valid but not enabled.

services/src/main/java/org/apache/unomi/services/impl/rules/RulesServiceImpl.java
```diff
@@ -454,6 +470,8 @@ public void setRule(Rule rule) {
     }
 
     public void removeRule(String ruleId) {
+        Set<Rule> newRules = allRules.stream().filter(rule -> !rule.getItemId().equals(ruleId)).collect(Collectors.toSet());
+        allRules = new ArrayList<>(newRules);
```
jkevan, line 474: Don't we risk concurrency issue in case of multiple update of this list ?
in case there is the job running in // or multiple threads using the list ?

jkevan (changes requested):

dgriffon:
Drop this PR to change the global approach to fix the issue. Disabling the rule was not working as it might be an issue to identify rules that have been disabled because of invalidity and rules willingly disabled. This mostly when we need to reenable them once conditions are met.

PR #368, opened 2021-12-20
UNOMI-537 Add control groups to personalizations

sergehuber:
- Personalizations will now store a control group election in the session or in the profile
- Updated documentation for personalization
- Implemented integration tests for control groups

### DIFF
--- a/rest/src/main/java/org/apache/unomi/rest/endpoints/ContextJsonEndpoint.java
+++ b/rest/src/main/java/org/apache/unomi/rest/endpoints/ContextJsonEndpoint.java
@@ -20,13 +20,7 @@ package org.apache.unomi.rest.endpoints;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.cxf.rs.security.cors.CrossOriginResourceSharing;
-import org.apache.unomi.api.ContextRequest;
-import org.apache.unomi.api.ContextResponse;
-import org.apache.unomi.api.Event;
-import org.apache.unomi.api.Persona;
-import org.apache.unomi.api.PersonaWithSessions;
-import org.apache.unomi.api.Profile;
-import org.apache.unomi.api.Session;
+import org.apache.unomi.api.*;
 import org.apache.unomi.api.conditions.Condition;
 import org.apache.unomi.api.services.ConfigSharingService;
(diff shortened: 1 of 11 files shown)


rest/src/main/java/org/apache/unomi/rest/endpoints/ContextJsonEndpoint.java
```diff
@@ -378,6 +372,26 @@ private Changes handleRequest(ContextRequest contextRequest, Session session, Pr
...
+        if (personalizations != null) {
+            data.setPersonalizations(new HashMap<>());
+            for (PersonalizationService.PersonalizationRequest personalization : sanitizePersonalizations(personalizations)) {
+                PersonalizationResult personalizationResult = personalizationService.personalizeList(profile, session, personalization);
+                changes.setChangeType(changes.getChangeType() | personalizationResult.getChangeType());
+                data.getPersonalizations()
+                        .put(personalization.getId(), personalizationResult.getContentIds());
+            }
+        }
+
```
jkevan, line 394: We could consider using event mechanism to resolve the personalizations, since it can now modify the profile or the session it would make sense.
And we would benefit from the "changes" return by the event evaluation instead of having to implement a changes handler inside the personalizaton layer.

We would need:
-  a new event: personalizationEvent that contains the perso(s) to be evaiuated
- a new action: resolve personalizations
- a new rules that listen on this event to resolve the perso

The only blocking point would be getting the result of the perso from this kind of execution and pass it to the contextResponse as it was done before.
May be we could store this in the event it self and catching those events here to get the result out of them.

I dont know why, but I think trying to rely the more we can on rules/events/actions for internal mechanism improve the genericity of unomi features.

jkevan (approved):

rest/src/main/java/org/apache/unomi/rest/endpoints/ContextJsonEndpoint.java
sergehuber: I like this suggestion. Should we merge this first and then rewrite it or should we change directly in the same PR ?

jkevan: Yes no problem for merging this first. And we will do the additional work in a separate PR as it was not the goal of the current PR.

PR #382, opened 2022-01-28
UNOMI-542 : fix integration tests

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-542


sergehuber (approved):
Thanks for the fix Jonathan, looks good to me. 
Regards,
  Serge...

PR #393, opened 2022-03-10
UNOMI-555 : add possiblity to store jsonSchema

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-555


sergehuber (changes requested):
Hello Jonathan, thanks a lot for this PR. I saw some minor things that need fixing before we can merge, I put comments in the files directly.

PR #403, opened 2022-04-13
UNOMI-487 Add profile alias management to REST API and GraphQL API

anatol-sialitski:
UNOMI-487 Add profile alias management to REST API and GraphQL API

In scope of this PR was implemented REST API and GraphQL API for profile aliases.

1. Added new ES index `profileAlias`
2. GraphQL API
3. REST API
4. Integration tests


api/src/main/java/org/apache/unomi/api/services/ProfileService.java
```diff
@@ -110,6 +110,10 @@ public interface ProfileService {
 
     void addAliasToProfile(String profileID, String alias, String clientID);
 
+    ProfileAlias removeAliasFromProfile(String profileID, String alias, String clientID);
```
sergehuber, line 113: Could you please add Javadoc for all the profile alias methods ? This is a requirement of the API classes

sergehuber (changes requested):
Thanks a lot for the PR @anatol-sialitski. I just noticed some missing Javadoc comments on the ProfileService. Could you add them so I can accept this PR?
Thanks, regards,
  Serge Huber.

api/src/main/java/org/apache/unomi/api/services/ProfileService.java
anatol-sialitski: done

sergehuber (approved):
Thanks a lot for the quick changes. Looks good to me now, I will merge it.

PR #408, opened 2022-04-26
UNOMI-561 : Add merge extension logic

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-561


api/src/main/java/org/apache/unomi/api/schema/JsonSchemaEntity.java
```diff
@@ -22,14 +22,14 @@
 /**
  * Object which represents a JSON schema stored in the persistence service
  */
-public class UnomiJSONSchema extends MetadataItem {
+public class JsonSchemaEntity extends MetadataItem {
```
dgriffon, line 25: please rename to `JSONSchemaEntity` to match naming format

services/src/main/java/org/apache/unomi/services/impl/schemas/SchemaServiceImpl.java
```diff
@@ -266,6 +272,69 @@ public JsonSchema getJsonSchema(String schemaId) {
...
+            JsonNode targetNode = schemaNode.at(path);
+            JsonNode nodeToAdd = entry.getValue();
+            if (targetNode.isArray()) {
+                handleArray((ArrayNode) targetNode, nodeToAdd);
+            } else if (targetNode.isObject() && nodeToAdd.isObject()) {
+                mergeIntoSchema(targetNode, nodeToAdd);
+            } else if (targetNode instanceof MissingNode) {
+                ((ObjectNode) schemaNode).set(path.replace("/", ""), nodeToAdd);
+            }
+        });
```
dgriffon, line 313: What happen if the json node is neither an Object nor an Array?

dgriffon (changes requested):

services/src/main/java/org/apache/unomi/services/impl/schemas/SchemaServiceImpl.java
jsinovassin: If it's another type, for example text, number, boolean, we don't change the value of the property in the schema. We simply skip, so the user can not override simple property.

dgriffon (approved):
Thank you for the contribution!

PR #413, opened 2022-05-03
[UNOMI-566] Allow deserialization of profile in event

aoliva-sefas:
Register Profile class in Item deserializer.


sergehuber:
Hello Adrien, 

Thanks for the PR, but after some thinking this might be problematic in terms of security. In Unomi 2 we have a new system to protect the data that allows these types to be used. 

Let's continue this discussion in the mailing list.

Regards,
  Serge...

sergehuber:
After further consideration this is too risky from a security standpoint so I prefer we don't include it.

PR #414, opened 2022-05-03
POC Extendable keyword

sergehuber:
Quick prototype proof of concept for an extendable JSON schema keyword to lookup all extensions for a given schema and evaluate them upon validation.


sergehuber:
This POC is no longer valid since we have completely refactored the schema service.

PR #420, opened 2022-05-10
remove property in login schema

jsinovassin:


sergehuber:
Hello Jonathan, thanks for the PR. 

Could you provide at least a ticket number in this PR?

Regards,
  Serge...

dgriffon (approved):
approved, please wait for one integration test to pass before merging

Also this schemas should not be part of Unomi (as well as the related event types). We need to move them where they are used (and created)

PR #421, opened 2022-05-10
UNOMI-568: refacto json-schema implementation

jkevan:
Big refactor for jsonSchema implementation.


extensions/json-schema/rest/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>osgi.cmpn</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
```
sergehuber, line 86: Do we need to embed this ?

extensions/json-schema/services/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+        <dependency>
+            <groupId>org.apache.unomi</groupId>
+            <artifactId>unomi-persistence-spi</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
```
sergehuber, line 69: Same comment here, is this required to be embedded ?

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/JsonSchemaWrapper.java
```diff
@@ -14,22 +14,20 @@
...
+package org.apache.unomi.schema.api;
 
 import org.apache.unomi.api.Metadata;
 import org.apache.unomi.api.MetadataItem;
 
-/**
- * Object which represents a JSON schema stored in the persistence service
- */
-public class UnomiJSONSchema extends MetadataItem {
+public class JsonSchemaWrapper extends MetadataItem {
```
sergehuber, line 23: Could you please add a small Javadoc description ?

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/SchemaService.java
```diff
@@ -100,56 +84,17 @@ public interface SchemaService {
...
-     * @param extensionStream inputStream of the extension to delete
-     * @return true if the extension has been deleted
+     * @param schemaStream inputStream of the schema
      */
-    boolean deleteExtension(InputStream extensionStream) throws IOException;
+    void loadPredefinedSchema(InputStream schemaStream) throws IOException;
 
     /**
-     * Delete an extension by an id
+     * unload a predefined schema into memory
```
sergehuber, line 94:
```suggestion
     * Unload a predefined schema into memory
```

sergehuber (changes requested):
Hello Kevan, 

Thanks for your PR, I found just a few minor issues, otherwise it looks good.

Best regards,
  Serge...

extensions/json-schema/services/pom.xml
jkevan: it's also the case in all other bundles in Unomi I dont think we need to embed it anywhere, but it will need to be change in a lot of places including this one. We will have to create a dedicated ticket for cleaning this in a global scale.

dgriffon (approved):

PR #424, opened 2022-05-17
UNOMI-395 : add endpoint to store scopes

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-395


services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+        for (Scope definition : items.getList()) {
+            details.add(definition.getMetadata());
+        }
+        return new PartialList<>(details, items.getOffset(), items.getPageSize(), items.getTotalSize(), items.getTotalSizeRelation());
+    }
+
+    @Override
+    public void save(Scope scope) {
+        if (persistenceService.save(scope)) {
+            persistenceService.refreshIndex(Scope.class, null);
```
jkevan, line 49: Why is there is manual refresh here ?

jkevan (approved):

services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+
+public class ScopeServiceImpl implements ScopeService {
+
+    private PersistenceService persistenceService;
+
+    private SchedulerService schedulerService;
+
+    private Integer scopesRefreshInterval = 1000;
+
+    private Map<String, Scope> scopes = new HashMap<>();
```
jkevan, line 44: Should be a concurrent map

services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        for (Scope definition : items.getList()) {
+            details.add(definition.getMetadata());
+        }
+        return new PartialList<>(details, items.getOffset(), items.getPageSize(), items.getTotalSize(), items.getTotalSizeRelation());
+    }
+
+    @Override
+    public void save(Scope scope) {
+        if (persistenceService.save(scope)) {
+            scopes.put(scope.getItemId(), scope);
```
jkevan, line 81: Nothing should change the map other than the refresh to avoid concurrency issues.

services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    @Override
+    public boolean delete(String id) {
+        if (persistenceService.remove(id, Scope.class)) {
+            scopes.remove(id);
```
jkevan, line 88: Nothing should change the map other than the refresh to avoid concurrency issues.

services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public Scope getScope(String id) {
+        return persistenceService.load(id, Scope.class);
```
jkevan, line 96: What the poiint of having in memory loaded stuff if we are reading the database here ?

jkevan (changes requested):

services/src/main/java/org/apache/unomi/services/impl/scope/ScopeServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public PartialList<Metadata> getScopesMetadatas(int offset, int size, String sortBy) {
+        PartialList<Scope> items = persistenceService.getAllItems(Scope.class, offset, size, sortBy);
+        List<Metadata> details = new LinkedList<>();
+        for (Scope definition : items.getList()) {
+            details.add(definition.getMetadata());
+        }
+        return new PartialList<>(details, items.getOffset(), items.getPageSize(), items.getTotalSize(), items.getTotalSizeRelation());
+    }
```
jkevan, line 76: Since we are loading the scopes in RAM, I'm not sure it's interesting to keep this query.
We should return the list of in RAM Scopes, witch make the parameters like: int offset, int size, String sortBy obsolete.

jkevan (changes requested):

jkevan (approved):

PR #427, opened 2022-05-20
UNOMI-395 : add migration logic to create scopes entries from scope i…

jsinovassin:
…n existing events

https://issues.apache.org/jira/browse/UNOMI-395

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
@@ -16,6 +16,7 @@
  */
 package org.apache.unomi.shell.migration.impl;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
@@ -33,6 +34,9 @@ import org.osgi.framework.Version;
 import org.osgi.service.component.annotations.Component;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
@@ -45,6 +49,7 @@ public class MigrationTo200 implements Migration {
     private CloseableHttpClient httpClient;
     private Session session;
     private String esAddress;
+    private BundleContext bundleContext;
 
     @Override
     public Version getFromVersion() {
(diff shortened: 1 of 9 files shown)


tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
```diff
@@ -126,25 +113,88 @@ private void copyValueScopeToSourceId(final String indexName, final CloseableHtt
...
+                return indexesAsJson.keySet().stream().
+                        filter(alias -> alias.startsWith(indexPrefix + "-event")).
+                        collect(Collectors.toSet());
+            }
+        }
+        return Collections.emptySet();
+    }
+
+    private void createScopes(Set<String> scopes, String indexPrefix) throws IOException {
+        final StringBuilder body = new StringBuilder();
```
jkevan, line 147: Careful here it seem's you forgot to create the scope index with the mapping provided in:
persistence-elasticsearch/core/src/main/resources/META-INF/cxs/mappings/scope.json

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
```diff
@@ -126,25 +113,88 @@ private void copyValueScopeToSourceId(final String indexName, final CloseableHtt
...
+
+        Set<String> scopes = new HashSet<>();
+        try (CloseableHttpResponse response = httpClient.execute(httpPost)) {
+            JSONObject responseAsJson = new JSONObject(EntityUtils.toString(response.getEntity()));
+            if (response.getStatusLine().getStatusCode() == HttpStatus.SC_OK) {
+                System.out.println("Getting the \"scope\" values from the events successfully finished. " + "Number of scope to create: "
+                        + responseAsJson.getJSONObject("aggregations").getJSONObject("bucketInfos").get("count").toString());
+                scopes = StreamSupport
+                        .stream(responseAsJson.getJSONObject("aggregations").getJSONObject("scopes").getJSONArray("buckets").spliterator(),
+                                false).map(bucketElement -> ((JSONObject) bucketElement).getString("key")).collect(Collectors.toSet());
```
jkevan, line 192: is it null safe ? if one of the element is null during the chain of functions will it break the entire migration ?
Migration code should be as safe as possible.

jkevan (changes requested):

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
jsinovassin: Yes, it's null safe an exception would be thrown if the attribute is not found and the migration will not crash

jkevan (approved):

PR #428, opened 2022-05-25
UNOMI-569 Flatten JSON Schemas & fix schema bugs

sergehuber:
- All schemas are now "flattened", meaning that each object is in a separate schema file (with a few exceptions)
- Bug fixes in GraphQLSchemaProvider to detect loops in schemas
- Fix events used in integration tests to avoid sending profile objects
- Changed Event default initialization to avoid creating attributes when using default constructor
- Added JAXBAnnotationModule in object mapper used in integration tests
- Added system property to control logging in integration tests (to make it easier to see invalid schema logs that exist only in debug)
- Fix deployment of test schemas in integration tests


extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/metadataitem.json
```diff
@@ -1,6 +1,12 @@
 {
   "$id": "https://unomi.apache.org/schemas/json/metadataitem/1-0-0",
   "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "self":{
+    "vendor":"The Apache Software Foundation",
+    "name": "metadataItem",
+    "format":"jsonschema",
+    "version":"1-0-0"
+  },
```
jkevan, line 9: This JSON Schema is not required and should be removed.

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/values/string.json
```diff
@@ -1,6 +1,13 @@
 {
     "$id": "https://unomi.apache.org/schemas/json/values/string/1-0-0",
     "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "self":{
+        "vendor":"The Apache Software Foundation",
+        "target" : "values",
+        "name": "string",
+        "format":"jsonschema",
+        "version":"1-0-0"
+    },
```
jkevan, line 10: do we still need this schemas on values type ?

extensions/web-tracker/wab/src/main/resources/META-INF/cxs/schemas/articleCompleted.json
```diff
@@ -2,23 +2,18 @@
   "$id": "https://unomi.apache.org/schemas/json/events/articleCompleted/1-0-0",
   "$schema": "https://json-schema.org/draft/2019-09/schema",
   "self":{
-    "vendor":"org.apache.unomi",
+    "vendor":"The Apache Software Foundation",
```
jkevan, line 5: Careful, this schema doesn't contains "unevaluatedProperty:false", so it could produce mapping explosion

extensions/web-tracker/wab/src/main/resources/META-INF/cxs/schemas/identify.json
```diff
@@ -2,7 +2,7 @@
   "$id": "https://unomi.apache.org/schemas/json/events/identify/1-0-0",
   "$schema": "https://json-schema.org/draft/2019-09/schema",
   "self":{
-    "vendor":"org.apache.unomi",
+    "vendor":"The Apache Software Foundation",
```
jkevan, line 5: Careful, this schema doesn't contains "unevaluatedProperty:false", so it could produce mapping explosion

extensions/web-tracker/wab/src/main/resources/META-INF/cxs/schemas/identify/properties.json
```diff
@@ -0,0 +1,24 @@
...
+    "nickname": {
+      "type": "string"
+    },
+    "favoriteCompiler": {
+      "type": "string"
+    },
+    "industry": {
+      "type": "string"
+    }
+  }
```
jkevan, line 23: Careful, this schema doesn't contains "unevaluatedProperty:false", so it could produce mapping explosion

extensions/web-tracker/wab/src/main/resources/META-INF/cxs/schemas/identify.json
```diff
@@ -13,12 +13,7 @@
...
-      "type" : "object",
-      "properties" : {
-        "consent" : {
-          "$ref" : "https://unomi.apache.org/schemas/json/consent/1-0-0"
-        }
-      }
+      "$ref" : "https://unomi.apache.org/schemas/json/events/identify/properties/1-0-0"
     },
     "source" : {
       "$ref" : "https://unomi.apache.org/schemas/json/customitems/page/1-0-0"
```
jkevan, line 19: Does this event should not have his own JSON Schemas for the target/source object ?

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/customitem.json
```diff
@@ -1,6 +1,12 @@
 {
   "$id": "https://unomi.apache.org/schemas/json/customitem/1-0-0",
   "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "self":{
+    "vendor":"The Apache Software Foundation",
+    "name": "customItem",
+    "format":"jsonschema",
+    "version":"1-0-0"
+  },
```
jkevan, line 9: Not sure this schema is still necessary, I dont see why event should use customItem for now.

jkevan (changes requested):

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/customitems/page.json
```diff
@@ -16,58 +16,9 @@
     }
   ],
```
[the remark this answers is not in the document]
sergehuber: If I do that I would have to add the itemId and itemType directly in the page object, is this what you are suggesting ?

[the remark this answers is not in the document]
sergehuber: I'm not sure if this is better than removing the custom item. Note that custom item here means the CustomItem class, not a custom item as something that is managed in its own index, but maybe the confusion is coming from there ?

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/customitems/page/properties/pageInfo.json
```diff
@@ -0,0 +1,58 @@
+{
+  "$id": "https://unomi.apache.org/schemas/json/customitems/page/properties/pageInfo/1-0-0",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "self": {
+    "vendor": "org.apache.unomi",
+    "name": "pageInfo",
+    "format": "jsonschema",
+    "version": "1-0-0"
+  },
```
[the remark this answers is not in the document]
sergehuber: This will not be handled in this PR but in a separate one.

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/events/form.json
```diff
@@ -2,7 +2,7 @@
   "$id": "https://unomi.apache.org/schemas/json/events/form/1-0-0",
   "$schema": "https://json-schema.org/draft/2019-09/schema",
   "self":{
-    "vendor":"org.apache.unomi",
+    "vendor":"The Apache Software Foundation",
```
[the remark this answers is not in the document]
sergehuber: Also, this will be handled in a separate issue.

jkevan (approved):

jkevan (changes requested):
I will contribute to this PR to clean a bit the existing schemas

extensions/web-tracker/wab/src/main/resources/META-INF/cxs/schemas/identify.json
sergehuber: In the example it is by default using the site as a source and a page as a target but in the schema it was not correct. I'll fix that.

itests/src/test/java/org/apache/unomi/itests/BasicIT.java
```diff
@@ -39,6 +39,7 @@
 import org.apache.unomi.api.services.RulesService;
 import org.apache.unomi.itests.tools.httpclient.HttpClientThatWaitsForUnomi;
 import org.apache.unomi.persistence.spi.CustomObjectMapper;
+import org.apache.unomi.schema.api.SchemaService;
```
jsinovassin, line 42: Seems that is not used in the test

jsinovassin (approved):

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/items/interests/interests.json
```diff
@@ -0,0 +1,18 @@
...
+  "self":{
+    "vendor":"org.apache.unomi",
+    "name": "interests",
+    "format":"jsonschema",
+    "version":"1-0-0"
+  },
+  "title": "Interests",
+  "type": "object",
+  "patternProperties": {
+    "^.*$": {
```
jsinovassin, line 13: Can't we limit to only alphanumerics characters?

sergehuber: This might cause problems in other languages such as Chinese, Cyrillic, etc...

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/items/page/page.properties.pageInfo.json
```diff
@@ -0,0 +1,63 @@
...
+      "type": ["null", "string"],
+      "format": "uri"
+    },
+    "tags": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "isContentTemplate": {
```
sergehuber, line 55: I was thinking we could use extensions for non-Unomi properties such as these, what do you think ?

sergehuber (commented):
Hi Kevan, 

Thanks for the hard work, it looks good on my side.

I just put a question in one comment.

Regards,
  Serge...

itests/src/test/java/org/apache/unomi/itests/BasicIT.java
jkevan: True, I will remove it. Thx !

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/items/interests/interests.json
jkevan: For now we will not restreint this, if validation is asked on the keys of the interests we will do something, for now it's not the case.

extensions/json-schema/services/src/main/resources/META-INF/cxs/schemas/items/page/page.properties.pageInfo.json
jkevan: This properties if necessary have to be identified and if require we will manage this in a separate PR once actions are defined on the subject.

PR #429, opened 2022-05-30
UNOMI-580 : add endpoint to get schema + clean

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-580

### DIFF
--- a/extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
+++ b/extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
@@ -27,10 +27,13 @@ import org.slf4j.LoggerFactory;
 
 import javax.jws.WebMethod;
 import javax.jws.WebService;
-import javax.ws.rs.*;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import java.util.Base64;
 import java.util.Set;
 
 @WebService
@@ -66,6 +69,18 @@ public class JsonSchemaEndPoint {
     }
 
     /**
+     * Get a schema by it's id
+     *
+     * @param id of the schema
+     * @return Json schema matching the id
+     */
+    @POST
+    @Path("/queryId")
+    public String getSchema(String id) {
+        return schemaService.getSchema(id).getSchema().replace("\n", "");
+    }
+
+    /**
      * Save a JSON schema
      *
      * @param jsonSchema the schema as string to save
@@ -73,7 +88,7 @@ public class JsonSchemaEndPoint {
      */
     @POST
     @Path("/")
-    @Consumes({MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON})
+    @Consumes({ MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON })
     @Produces(MediaType.APPLICATION_JSON)
     public Response save(String jsonSchema) {
         try {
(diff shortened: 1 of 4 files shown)


extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
```diff
@@ -65,6 +68,18 @@ public Set<String> getInstalledJsonSchemaIds() {
...
     }
 
+    /**
+     * Get a schema by it's id
+     *
+     * @param id of the schema
+     * @return Json schema matching the id
+     */
+    @POST
+    @Path("/queryId")
```
jkevan, line 78: may be we could name it "query" directly ? like what we have for the "delete"

itests/src/test/java/org/apache/unomi/itests/JSONSchemaIT.java
```diff
@@ -64,155 +72,176 @@ public void setUp() throws InterruptedException {
...
-    public void testEndPoint_SaveDelete() throws InterruptedException, IOException {
+    public void testEndPoint_GetJsonSchemasById() throws Exception {
+        // Push base schemas
+        schemaService.saveSchema(resourceAsString("schemas/schema-dummy.json"));
+        schemaService.saveSchema(resourceAsString("schemas/schema-dummy-properties.json"));
+
+        final String schemaId = "https://unomi.apache.org/schemas/json/events/dummy/1-0-0";
+        final HttpPost request = new HttpPost(getFullUrl(JSONSCHEMA_URL + "/queryId"));
+
+        request.setEntity(new StringEntity(schemaId, ContentType.APPLICATION_JSON));
```
jkevan, line 211: This is not JSON Actually, we should may not use this content type here.

itests/src/test/java/org/apache/unomi/itests/JSONSchemaIT.java
```diff
@@ -64,155 +72,176 @@ public void setUp() throws InterruptedException {
...
+        keepTrying("Schema should have been created",
+                () -> schemaService.getSchema("https://unomi.apache.org/schemas/json/events/dummy/1-0-0"), Objects::nonNull,
+                DEFAULT_TRYING_TIMEOUT, DEFAULT_TRYING_TRIES);
 
         // Delete Schema using REST call
-        String encodedString = Base64.getEncoder()
-                .encodeToString("https://unomi.apache.org/schemas/json/events/dummy/1-0-0".getBytes());
-        CloseableHttpResponse response = delete(JSONSCHEMA_URL + "/" + encodedString);
+        final HttpPost request = new HttpPost(getFullUrl(JSONSCHEMA_URL + "/delete"));
+        request.setEntity(new StringEntity("https://unomi.apache.org/schemas/json/events/dummy/1-0-0", ContentType.APPLICATION_JSON));
```
jkevan, line 239: This is not JSON Actually, we should may not use this content type here.

extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
```diff
@@ -85,14 +100,13 @@ public Response save(String jsonSchema) {
...
      */
-    @DELETE
-    @Path("/{base64JsonSchemaId}")
-    public void remove(@PathParam("base64JsonSchemaId") String base64JsonSchemaId) {
-        schemaService.deleteSchema(new String(Base64.getDecoder().decode(base64JsonSchemaId)));
+    @POST
+    @Path("/delete")
+    public void remove(String id) {
+        schemaService.deleteSchema(id);
     }
```
jkevan, line 111: Could be great to have a response here, in case the id doest exist for exemple, or even if the id is matching a predefined schema, in this case the schemaService.deleteSchema is returning false. 
So it could be great to reflect the status of the service return in the endPoint.

extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
```diff
@@ -65,6 +68,18 @@ public Set<String> getInstalledJsonSchemaIds() {
...
+    /**
+     * Get a schema by it's id
+     *
+     * @param id of the schema
+     * @return Json schema matching the id
+     */
+    @POST
+    @Path("/queryId")
+    public String getSchema(String id) {
+        return schemaService.getSchema(id).getSchema().replace("\n", "");
```
jkevan, line 80: be careful of NPE here what happen if: schemaService.getSchema(id) is returning null ? :)

jkevan (changes requested):

jkevan (approved):

jkevan (approved):

PR #440, opened 2022-06-15
UNOMI-505 Study replication of existing profileIDs into new alias index

anatol-sialitski:
Added the migration which creates a document in `context-profilealias` index based on the profile.

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
@@ -19,6 +19,7 @@ package org.apache.unomi.shell.migration.impl;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpPut;
@@ -28,6 +29,7 @@ import org.apache.http.util.EntityUtils;
 import org.apache.karaf.shell.api.console.Session;
 import org.apache.unomi.shell.migration.Migration;
 import org.apache.unomi.shell.migration.utils.ConsoleUtils;
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Version;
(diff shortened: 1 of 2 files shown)


tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
```diff
@@ -246,6 +253,109 @@ private Set<String> getSetOfScopes(Set<String> indices) throws IOException {
...
+    private void createProfileAliasDocumentsFromProfile() throws IOException {
+        System.out.println("Migration \"Create profileAlias from profile\" started");
+        Instant migrationTime = Instant.now();
+        int initialOffset = 1000;
+        int size = 1000;
+        doProcessResponse(migrationTime, initialOffset, size, null);
+        System.out.println("Migration \"Create profileAlias from profile\" completed.");
+    }
+
+    private void doProcessResponse(Instant migrationTime, int initialOffset, int offset, String scrollId) throws IOException {
```
sergehuber, line 265: I think we should rename this method to something like processProfiles or something like that, because we are not processing a response here.

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/impl/MigrationTo200.java
```diff
@@ -246,6 +253,109 @@ private Set<String> getSetOfScopes(Set<String> indices) throws IOException {
...
+                        }
+
+                        CloseableHttpResponse bulkResponse = httpClient.execute(createProfileAliasRequest(bulkCreateRequest.toString()));
+                        if (bulkResponse != null) {
+                            bulkResponse.close();
+                        }
+                    }
+
+                    if (hitsObject.getJSONObject("total").getInt("value") > offset) {
+                        doProcessResponse(migrationTime, initialOffset, offset + initialOffset, responseAsJson.getString("_scroll_id"));
```
sergehuber, line 300: I don't think we should do a recursive call here. After all we are using a scroll query so we should simply use a continue the scroll query until all profiles have been processed. Also, the use of a recursive call will be very problematic if we have million of profiles in the data, which can happen.

sergehuber: Also using a continue scroll query will make the handling of the result count easier.

sergehuber (changes requested):
Hello @anatol-sialitski ,

Thank for your PR. I found some small issues with it, I put the comments in the code directly. Don't hesitate if you have any questions.

Best regards,
  Serge Huber.

sergehuber (approved):
Hi Anatol, 

Thanks for implemented the requested changes. Looks good to me now !

Regards,
  Serge...

PR #441, opened 2022-06-17
UNOMI-597: do not persist internal events

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-597

### DIFF
--- a/plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/SendEventAction.java
+++ b/plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/SendEventAction.java
@@ -17,6 +17,7 @@
 
 package org.apache.unomi.plugins.baseplugin.actions;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.unomi.api.Event;
 import org.apache.unomi.api.Item;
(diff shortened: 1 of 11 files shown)


plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/actions/SendEventAction.java
```diff
@@ -36,18 +37,19 @@ public void setEventService(EventService eventService) {
...
-//        String type = (String) target.get("type");
-
-//            Item targetItem = new CustomItem();
-//            BeanUtils.populate(targetItem, target);
 
         Event subEvent = new Event(eventType, event.getSession(), event.getProfile(), event.getSourceId(), event, target, event.getTimeStamp());
         subEvent.setProfileId(event.getProfileId());
         subEvent.getAttributes().putAll(event.getAttributes());
         subEvent.getProperties().putAll(eventProperties);
+        if(!toBePersisted){
```
jkevan, line 50: Missing null check, this parameter was not existing before, there is a chance that is will be null.
If null we should continue to have the same behavior as before for compatibility reasons.

jkevan (changes requested):

jkevan (approved):

PR #442, opened 2022-06-17
UNOMI-585: introduce new nestedCondition and set profile.properties.interests as nested

jkevan:


itests/src/test/java/org/apache/unomi/itests/ConditionEvaluatorIT.java
```diff
@@ -47,6 +47,7 @@
 public class ConditionEvaluatorIT extends BaseIT {
     protected ConditionBuilder builder;
     protected Item item;
+    protected Item emptyItem;
```
dgriffon, line 50: Can it be renamed to emptyProfile?

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/NestedConditionEvaluator.java
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+
+        return flattenNestedItem;
+    }
+
+    protected Item createFinalNestedItemForEvaluation(Item parentItem, String path, Map<String, Object> flattenedNestedItem) {
+        // Build a basic Item with merged properties
+        if (parentItem instanceof Profile) {
```
dgriffon, line 109: What about the other kind of item (like customItems) ?

dgriffon (approved):

plugins/baseplugin/src/main/java/org/apache/unomi/plugins/baseplugin/conditions/NestedConditionEvaluator.java
jkevan: they cannot provide ES Mapping anyway, so they cannot use nested ES mapping. 
So for now I would say it's not in the current scope.

PR #443, opened 2022-06-17
[UNOMI-601] - Integration of timeStamp field to metadata class

Saadmairaj:
Added a timestamp attribute which defines and stores the current time when the Metadata class is initialised.

This can be very useful as It gets challenging to sort and manage documents like segments, campaigns, lists, goals & rules as the timestamp field is missing. Adding an extra field timeStamp to metadata will give a timestamp field for all the documents other than profiles, sessions & events.

### DIFF
--- a/api/src/main/java/org/apache/unomi/api/Metadata.java
+++ b/api/src/main/java/org/apache/unomi/api/Metadata.java
@@ -20,6 +20,7 @@ package org.apache.unomi.api;
 import java.io.Serializable;
 import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.Date;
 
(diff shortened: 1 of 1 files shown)


sergehuber:
Hello, 

Thank you for your interesting contribution. Please note that we require integration tests for such contributions. Also, I've put a comment on the ticket (https://issues.apache.org/jira/browse/UNOMI-601) because I was thinking about these type of metadata more globally. Could you please check and look there to see how we want to move forward with this?

Best regards,
  Serge Huber.

sergehuber:
Can we close this PR as there hasn't been any reply since June 25th ?

PR #445, opened 2022-06-21
UNOMI-607: missing view properties schema

jkevan:


sergehuber (approved):
Thanks Kevan. 

Looks good to me.

PR #448, opened 2022-06-27
UNOMI-586: Optimize ES mappings by ignoring conditions and Objects th…

jkevan:
…at are not used for queries


sergehuber (approved):
Hello Kevan, 

Thanks for the PR. These changes make sense. All good to me.

Regards,
  Serge...

PR #461, opened 2022-07-12
Unomi 603 migration profiles

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-603


tools/shell-commands/src/main/resources/requestBody/2.0.0/update_profile_interests.json
```diff
@@ -0,0 +1,4 @@
+{
+    "source": "List interests = new ArrayList(); if(ctx._source.properties.interests != null && !(ctx._source.properties.interests instanceof List)){ for (entry in ctx._source.properties.interests.entrySet()){ Map interest = new HashMap(); interest.put('key',entry.getKey()); interest.put('value', entry.getValue()); interests.add(interest);} ctx._source.properties.remove('interests'); ctx._source.properties.put('interests',interests);}",
+    "lang": "painless"
+}
```
jkevan, line 4: It could be interesting to help on dev + review painless scripts that the script it self should be the file content.
like: update_profile_interests.painless.

Even if IDE is not capable of handling those file, may be one day it will.
Also it would help a lot to benefit a least from indentation to be able to read/debug thoses scripts.

Also the "source" and "lang" could be created on the fly by the reindex function.

jkevan (changes requested):

tools/shell-commands/src/main/resources/requestBody/2.0.0/update_profile_interests.json
sergehuber: Funny I had the exact same idea :)

PR #464, opened 2022-07-26
UNOMI-203, UNOMI-632, UNOMI-633, UNOMI-634: improve integrations test with possibility to test migrations

jkevan:
Content of the PR:
- Silent migration feature, using cmd like: `migrate 1.6.0 true`, true is for silent (UNOMI-633)
- Configurable migration using new OSGI conf: org.apache.unomi.migration.cfg (UNOMI-633)
- Improve reindexation in migrations to reuse mappings configuration provided by Unomi bundles instead of duplicating them. (UNOMI-634)
- Improve integration to use KarafTestSupport from karaf libraries: (UNOMI-203)
- Improve integration tests to be able to test migrations (UNOMI-632)
- Implement integration test for migration from 1.6.x to actually 2.0.0 (UNOMI-632)
- Migrate remaining indices that need a reindexation by generifying the segment migration script to handle others indices in the same situation like: campaigns, goals, rules, condition types, etc ... (UNOMI-632)


sergehuber (approved):
Hello Kevan, 

Thanks a lot for doing this, it is something I've been wanting to have for a long time.
Looks all good to me.

Regards,
  Serge...

dgriffon (approved):

PR #469, opened 2022-08-02
UNOMI-637 : do not create mapping for property type if not necessary

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-637

### DIFF
--- a/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
+++ b/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
@@ -1438,16 +1438,19 @@ public class ElasticSearchPersistenceServiceImpl implements PersistenceService, 
 
     public void setPropertyMapping(final PropertyType property, final String itemType) {
         try {
+            Map<String, Object> propertyMapping = createPropertyMapping(property);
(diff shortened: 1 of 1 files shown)


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -1444,7 +1444,10 @@ public void setPropertyMapping(final PropertyType property, final String itemTyp
             }
             Map<String, Object> subMappings = mappings.computeIfAbsent("properties", k -> new HashMap<>());
             Map<String, Object> subSubMappings = (Map<String, Object>) subMappings.computeIfAbsent("properties", k -> new HashMap<>());
-            mergePropertiesMapping(subSubMappings, createPropertyMapping(property));
+            Map<String, Object> propertyMapping = createPropertyMapping(property);
+            if (propertyMapping != null) {
+                mergePropertiesMapping(subSubMappings, propertyMapping);
+            }
```
jkevan, line 1450: I think we should simplify this function, if `Map<String, Object> propertyMapping = createPropertyMapping(property);` is null we can simply abort this function by `return;`

Also for better perf, we can do this in the first line to avoid calculating the existing mapping in case createPropertyMapping return null.

jkevan (changes requested):

jkevan (approved):

PR #476, opened 2022-08-18
UNOMI-640: safer way of retreiving index config in migration

jkevan:


sergehuber (approved):
Thanks Kevan for the improvements, this should make the migration process a lot more reliable.

I have validated these changes, feel free to merge them when ready.

Regards,
  Serge...

PR #480, opened 2022-08-25
UNOMI-623 : store ids and types of events in session when it's created

jsinovassin:
https://jira.jahia.org/browse/UNOMI-623


sergehuber (approved):
Hello Jonathan, 

Thanks for the PR this looks good to me.

Regards,
  Serge...

PR #511, opened 2022-09-23
UNOMI-677: Add rule.actions.parameterValues open map to the ignored i…

jkevan:
…ndexed data in rules.json mapping


sergehuber (approved):
Hi Kevan,

Thanks for catching this and fixing it so quickly.

Looks good to me.

Regards,
  Serge...

PR #530, opened 2022-11-01
UNOMI-708 Invalid "no matching query builder log"

sergehuber:
- Only display log if we really couldn't find a query builder and not anymore in the case where we couldn't match parameterized values.

### DIFF
--- a/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/conditions/ConditionESQueryBuilderDispatcher.java
+++ b/persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/conditions/ConditionESQueryBuilderDispatcher.java
@@ -83,12 +83,12 @@ public class ConditionESQueryBuilderDispatcher {
             if (contextualCondition != null) {
                 return queryBuilder.buildQuery(contextualCondition, context, this);
             }
-        }
-
-        // if no matching
-        logger.warn("No matching query builder. See debug log level for more information");
-        if (logger.isDebugEnabled()) {
-            logger.debug("No matching query builder for condition {} and context {}", condition, context);
+        } else {
+            // if no matching
+            logger.warn("No matching query builder. See debug log level for more information");
+            if (logger.isDebugEnabled()) {
+                logger.debug("No matching query builder for condition {} and context {}", condition, context);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/conditions/ConditionESQueryBuilderDispatcher.java
```diff
@@ -83,12 +83,12 @@ public QueryBuilder buildFilter(Condition condition, Map<String, Object> context
             if (contextualCondition != null) {
                 return queryBuilder.buildQuery(contextualCondition, context, this);
             }
-        }
-
-        // if no matching
-        logger.warn("No matching query builder. See debug log level for more information");
-        if (logger.isDebugEnabled()) {
-            logger.debug("No matching query builder for condition {} and context {}", condition, context);
+        } else {
```
dgriffon, line 86: Don't we want to add a log when no matching condition happen too?

dgriffon (approved):

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/conditions/ConditionESQueryBuilderDispatcher.java
sergehuber: You mean when we can't resolve the parameters ? No I don't think so that would happen too often.

dgriffon: Indeed, I was talking about parameters, sounds like no matching signature in that case isn't it?

sergehuber: It's just that the event doesn't have a value for a parameter, which can happen on a lot of events. I think it might be too much noise to add a log for that.

dgriffon: ok, thanks for the answers !

dgriffon (approved):

PR #531, opened 2022-11-08
UNOMI-690, UNOMI-696: refactor control group

jkevan:
The goal is to clean and refactor the control group implementation to be able:
- correctly persist the status of the control group in the session/profile even when control group is false
- improve the response for personalization results to ease client side usage of this new informations

Now the control group are stored in the profile or session system properties, in the format:
```
{
  "systemProperties": {
    "personalizationStrategyStatus": [{
      "personalizationId": "40a6bca-348a-40c2-a00d-48c4346e8e2c",
      "inControlGroup": false,
      "timeStamp": "2022-11-08" 
    }]
  }
}
```
This way we can persist personalization strategy status on profile or session in a generic way that could be reuse for other strategies than control groups.

The control group will also be returned in the personalization results in the context response, using this format:
Before:
```
{
  "personalizations": {
    "40a6bca-348a-40c2-a00d-48c4346e8e2c": ["variant_1_id", "variant_2_id"]
  }
}
```

Now (We keep old syntax for retro-compatibility):
```
{
  "personalizations": {
    "40a6bca-348a-40c2-a00d-48c4346e8e2c": ["variant_1_id", "variant_2_id"]
  },
  "personalizationResults": {
    "40a6bca-348a-40c2-a00d-48c4346e8e2c": {
      "contentIds": ["variant_1_id", "variant_2_id"],
      "additionalResultInfos": {
        "inControlGroup": true
      }
    }
  }
}
```

This way we can return additional informations related to the personalized results that will be useful for the client side.


sergehuber (approved):
Hi Kevan, 

Thanks for all the work, including integration tests and refactoring. This make a lot more sense now. 

Looks good to me ! 

Regards,
  Serge...

PR #533, opened 2022-11-10
UNOMI-690, UNOMI-696: Fix java doc generation

jkevan:


sergehuber (approved):
Hi Kevan, 

Thanks for the quick fix, let's hope that does the trick :) 

Regards,
  Serge...

PR #542, opened 2022-11-16
UNOMI-710: revert Yauaa 7.8.0 upgrade

jkevan:
Due to Unomi build being broken due to the upgrade we will post pone the upgrade to next version to be able to release unomi 1.7.0


sergehuber (approved):
Hello Kevan,

Thanks for doing this revert so fast. Looks good to me. 

Regards,
  Serge...

PR #573, opened 2023-02-03
UNOMI-736: merge all item indices into same index

jkevan:


jkevan:
duplicate https://github.com/apache/unomi/pull/581

PR #575, opened 2023-02-08
Unomi 728 index migration

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-728


tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-2.2.0-10-rolloverAndMigrateSession.groovy
```diff
@@ -0,0 +1,74 @@
...
+            .replace("#pollIntervalValue", "\"2s\"")
+    HttpUtils.executePutRequest(context.getHttpClient(), esAddress + "/_cluster/settings", updatePollIntervalBody, null)
+})
+
+context.performMigrationStep("2.2.0-10-create-rollover-policy", () -> {
+    String createRolloverPolicyQuery = MigrationUtils.resourceAsString(bundleContext, "requestBody/2.2.0/create_rollover_policy_query.json")
+    String rolloverQueryBody = MigrationUtils.buildRolloverPolicyCreationRequest(createRolloverPolicyQuery, context)
+
+    HttpUtils.executePutRequest(context.getHttpClient(), esAddress + "/_ilm/policy/" + rolloverPolicyName, rolloverQueryBody, null)
+})
```
jkevan, line 41: There is no need to create the ILM twice, if the event migration already did it.
It's shared by both sessions and events. (actually it could make sense to handle session and events in the same migration script)

tools/shell-commands/src/main/resources/requestBody/2.2.0/base_index_mapping.json
```diff
@@ -0,0 +1,30 @@
+{
```
jkevan, line 1: Could be rename to base_index_withRollover_request.json ?

tools/shell-commands/src/main/resources/requestBody/2.2.0/create_rollover_policy_query.json
```diff
@@ -0,0 +1,19 @@
...
+            #rolloverHotActions
+          }
+        }
+      },
+      "delete": {
+        "min_age": "90d",
+        "actions": {
+          "delete": {}
+        }
+      }
```
jkevan, line 16: We really need to not forget about that during the purge story

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-2.2.0-10-rolloverAndMigrateSession.groovy
```diff
@@ -0,0 +1,74 @@
...
+    sortedIndices.each { sessionIndex ->
+        MigrationUtils.deleteIndex(context.getHttpClient(), esAddress, sessionIndex)
+    }
+})
+
+context.performMigrationStep("2.2.0-10-reset-poll-interval", () -> {
+    String updatePollIntervalBody = MigrationUtils.resourceAsString(bundleContext, "requestBody/2.2.0/update_settings_poll_interval.json")
+            .replace("#pollIntervalValue", "null")
+    HttpUtils.executePutRequest(context.getHttpClient(), esAddress + "/_cluster/settings", updatePollIntervalBody, null)
+})
```
jkevan, line 74: Could be done once if sessions/events are handle in same migration script.

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-2.2.0-10-rolloverAndMigrateSession.groovy
```diff
@@ -0,0 +1,74 @@
...
+Set<String> sessionIndices = MigrationUtils.getIndexesPrefixedBy(context.getHttpClient(), esAddress, indexPrefix + "-session-date-")
+List<String> sortedIndices = new ArrayList<>(sessionIndices)
+Collections.sort(sortedIndices)
+
+context.performMigrationStep("2.2.0-10-migrate-existing-sessions", () -> {
+    sortedIndices.each { sessionIndex ->
+        MigrationUtils.moveToIndex(context.getHttpClient(), bundleContext, esAddress, sessionIndex, indexPrefix + "-session")
+        sleep(3000)
+    }
+})
```
jkevan, line 62: if something fail in this step it will retry and potentially move again items that have been already moved.
I think we need a safer retry strategy on this overall reindex steps.
at the begining of this step we could delete by query all sessions/events in the new indices: context-session-00*

itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
```diff
@@ -84,8 +84,16 @@ public void checkMigratedData() throws Exception {
         checkViewEventRestructured();
         checkEventTypesNotPersistedAnymore();
         checkForMappingUpdates();
+        checkNewIndexesExists();
```
jkevan, line 87: would be greate to have more info in the method name: checkEventSessionRollover2_2_0 something like this

itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
```diff
@@ -84,8 +84,16 @@ public void checkMigratedData() throws Exception {
...
     }
 
+    /**
+     * Checks if at least the new index for events and sessions exists.
+     */
+    private void checkNewIndexesExists() throws IOException {
+        Assert.assertTrue(MigrationUtils.indexExists(httpClient, "http://localhost:9400", "context-event-000001"));
+        Assert.assertTrue(MigrationUtils.indexExists(httpClient, "http://localhost:9400", "context-session-000001"));
+    }
```
jkevan, line 96: Would be great to check that the number of sessions/events is same as before the migration.
We could do a count query on context-session-* and context-event-* before starting the migration and compare after the migration
Also we could check that old indices do not exists anymore: context-session-date-* and same for events.

jkevan (changes requested):
Is the snapshot for the migration containing:
- multiple months for sessions and/or events ?
- duplicate sessions/events ? in the previous implem it was possible to have same session id in multiple indices, we need to make sure the migration is proof to this corruption and will not fail in case it exists in the data to migrate.

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-2.2.0-10-rolloverAndMigrateSession.groovy
jsinovassin: Done

tools/shell-commands/src/main/resources/requestBody/2.2.0/base_index_mapping.json
jsinovassin: Done

tools/shell-commands/src/main/resources/requestBody/2.2.0/create_rollover_policy_query.json
jsinovassin: I added a line in Handle TODOs in the https://issues.apache.org/jira/browse/UNOMI-739 ticket to not forget it

jsinovassin:
> Is the snapshot for the migration containing:
> 
> * multiple months for sessions and/or events ?
> * duplicate sessions/events ? in the previous implem it was possible to have same session id in multiple indices, we need to make sure the migration is proof to this corruption and will not fail in case it exists in the data to migrate.

These is multiple months for sessions and events in the last snapshot, there is also some duplicate session in the indexes. 
The migration is not failing if there is some document with the same id in multiple index. But, at the end there will be only one document if the migration stores them in the same index

jkevan (approved):

jkevan (approved):

PR #578, opened 2023-02-21
Unomi 739 update purge system

jsinovassin:
https://jira.jahia.org/browse/UNOMI-739


jkevan (changes requested):

jkevan (approved):
It would be wonderful if we could have a new integration test to check that the empty indices are correctly removed during purge.
(may be no easy to test since you would need multiple rollover indices)

PR #581, opened 2023-02-24
UNOMI-736: merge all item indices into same index

jkevan:


extensions/groovy-actions/services/src/main/java/org/apache/unomi/groovy/actions/services/impl/GroovyActionsServiceImpl.java
```diff
@@ -70,41 +79,30 @@ public class GroovyActionsServiceImpl implements GroovyActionsService {
...
+    @Reference
     public void setPersistenceService(PersistenceService persistenceService) {
         this.persistenceService = persistenceService;
     }
 
-    public void setGroovyActionsRefreshInterval(Integer groovyActionsRefreshInterval) {
-        this.groovyActionsRefreshInterval = groovyActionsRefreshInterval;
-    }
-
+    @Reference()
```
jsinovassin, line 100: You can remove the parenthesis here.

itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
```diff
@@ -40,6 +40,7 @@ public class Migrate16xTo220IT extends BaseIT {
     private int sessionCount = 0;
 
     private static final int NUMBER_DUPLICATE_SESSIONS = 3;
+    private static final int NUMBER_PERSONASESSIONS = 2;
```
jsinovassin, line 43: should be NUMBER_PERSONA_SESSIONS

jsinovassin (approved):

PR #582, opened 2023-03-06
UNOMI-737: indices reduction migration (ES document id suffixed to av…

jkevan:
…oid conflicts)


sergehuber (approved):
Hi Kevan thanks for the PR, this looks a lot better than the previous code ! Looks good to me I approved it !

PR #584, opened 2023-03-09
UNOMI-700 : specify indices when doing a refresh

jsinovassin:
https://jira.jahia.org/browse/UNOMI-700


services/src/main/java/org/apache/unomi/services/impl/goals/GoalsServiceImpl.java
```diff
@@ -540,7 +540,7 @@ public GoalReport getGoalReport(String goalId, AggregateQuery query) {
     @Override
     public PartialList<CampaignEvent> getEvents(Query query) {
         if(query.isForceRefresh()){
-            persistenceService.refresh();
+            persistenceService.refreshIndex(Campaign.class);
```
jkevan, line 543: I think here it's CampaignEvent.class that need to be refreshed

jkevan (approved):

PR #591, opened 2023-03-17
UNOMI-752 : ensure target indices and alias does not exist before per…

dgriffon:
see https://issues.apache.org/jira/browse/UNOMI-752
perform a check on index and alias names before doing the migration to 2.2.0 of contents


jkevan (approved):
Thank you for the new checks for the migration. 
I just found the error message not easy to understand, the difference between Alias or Index is weird. 
If the index exists, it could be an index or an alias, but the message will say: Alias context-session should be removed. ( even if actually it's an index that exists and not an alias) 
May be a message like this would me more clear: "Migration requires to create a new index/alias NAME, but it seem's that this NAME is already used, please remove it to be able to continue the migration process"

But apart from that it's all good

PR #593, opened 2023-03-20
UNOMI-748, UNOMI-749: more improvements for merge system: updateWithS…

jkevan:
…cript to reassign past sessions/events, fix inconstency in sessions profileId, use schedulerService to support asynchronous update of past browsing data.


services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
```diff
@@ -163,7 +158,6 @@ public void postConstruct() throws IOException {
         }
         bundleContext.addBundleListener(this);
         initializeTimer();
-        loadScripts();
```
dgriffon, line 166: Is it related to the current PR?

dgriffon (approved):

services/src/main/java/org/apache/unomi/services/impl/segments/SegmentServiceImpl.java
jkevan: Yes I moved the painless script loading in the persistence service :)

PR #597, opened 2023-03-22
UNOMI-754: new jsonSchema endpoint to be able to validate events and …

jkevan:
…get validation error messages

### DIFF
--- a/extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
+++ b/extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
@@ -21,6 +21,7 @@ import org.apache.cxf.rs.security.cors.CrossOriginResourceSharing;
 import org.apache.unomi.rest.exception.InvalidRequestException;
 import org.apache.unomi.schema.api.JsonSchemaWrapper;
 import org.apache.unomi.schema.api.SchemaService;
+import org.apache.unomi.schema.api.ValidationMessageWrapper;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 import org.slf4j.Logger;
@@ -35,6 +36,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.util.Collection;
 import java.util.Set;
 
 @WebService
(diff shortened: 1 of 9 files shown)


extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/SchemaService.java
```diff
@@ -29,21 +29,39 @@ public interface SchemaService {
...
+
+    /**
+     * perform a validation on the given event
+     *
+     * @param event the event to validate
      * @return true is the event is valid
+     * @throws ValidationException in case something goes wrong and validation could not be performed.
      */
-    boolean isEventValid(String event, String eventType);
+    Set<ValidationError> validateEvent(String event) throws ValidationException;
```
dgriffon, line 64: The return type does not match the description.

extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
```diff
@@ -114,4 +116,22 @@ public Response save(String jsonSchema) {
...
+     * @param event the event to be validated
+     * @return Validation error messages if there is some
+     */
+    @POST
+    @Produces(MediaType.APPLICATION_JSON + ";charset=UTF-8")
+    @Consumes({ MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON })
+    @Path("/validateEvent")
+    public Collection<ValidationError> validateEvent(String event) {
+        try {
+            return schemaService.validateEvent(event);
```
dgriffon, line 131: There is small inconsistency in this method, we should return an error code always when the schema is invalid.
Ideally, all the errors should be handled by the Exception (if possible and have `schemaService.validateEvent()` to return a boolean. The `Collection<ValidationError>` can be carried by the ValidateException. 
If we don't want to do such change, at least we should return an error code in case of invalid event.

dgriffon (changes requested):

extensions/json-schema/rest/src/main/java/org/apache/unomi/schema/rest/JsonSchemaEndPoint.java
jkevan: I'm not sure, the goal of this endpoint is to return the validation errors.
So even if there is some, the request is valid and should be a 200.
The response code will be impacted if something goes wrong before doing the validation or in case of unexpected error during the validation process, that's why there is a difference between the exception and the validation errors.
The separation have been done on purpose.

ValidationException: Something unexpected happen and the validation was not able to be performed
ValidationError: The validation process is done successfully and we have informations on why the event does not respect the schemas.

dgriffon (approved):

PR #612, opened 2023-04-28
Unomi 775 add validation endpoint

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-775

### DIFF
--- a/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/ValidationError.java
+++ b/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/ValidationError.java
@@ -42,7 +42,8 @@ public class ValidationError implements Serializable {
     }
 
     public boolean equals(Object o) {
-        return validationMessage.equals(o);
+        ValidationError other = (ValidationError) o;
+        return validationMessage.equals(other.validationMessage);
     }
 
     public int hashCode() {
--- a/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/ValidationError.java
+++ b/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/ValidationError.java
@@ -27,23 +27,19 @@ import java.io.Serializable;
  */
 public class ValidationError implements Serializable {
 
-    private transient final ValidationMessage validationMessage;
+    private transient final String validationMessage;
 
-    public ValidationError(ValidationMessage validationMessage) {
+    public ValidationError(String validationMessage) {
         this.validationMessage = validationMessage;
     }
 
     public String getError() {
-        return validationMessage.getMessage();
-    }
-
-    public String toString() {
-        return validationMessage.toString();
+        return validationMessage;
     }
 
     public boolean equals(Object o) {
         ValidationError other = (ValidationError) o;
-        return validationMessage.equals(other.validationMessage);
+        return validationMessage.equals(other.getError());
     }
 
     public int hashCode() {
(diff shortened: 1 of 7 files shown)


extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
```diff
@@ -111,12 +112,52 @@ public boolean isEventValid(String event) {
...
+            } catch (ValidationException e) {
+                Set<ValidationError> errors = buildCustomErrorMessage();
+                if (errorsPerEventType.containsKey(eventType)) {
+                    errorsPerEventType.get(eventType).addAll(errors);
+                } else {
+                    errorsPerEventType.put(eventType, errors);
+                }
+                errorsPerEventType.put(eventType, errors);
+
+                logger.debug(e.getMessage());
```
jkevan, line 140: I think this two lines need to be removed ?

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/api/ValidationError.java
```diff
@@ -42,7 +42,8 @@ public String toString() {
     }
 
     public boolean equals(Object o) {
-        return validationMessage.equals(o);
+        ValidationError other = (ValidationError) o;
+        return validationMessage.equals(other.validationMessage);
```
jkevan, line 46: Could be simpler to replace directly the inner ValidationMessage object by the String error message at this step.
I would also simplify the building of custom error messages.

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
```diff
@@ -111,12 +112,52 @@ public boolean isEventValid(String event) {
...
-        String eventType = extractEventType(jsonEvent);
+        return validateNodeEvent(parseData(event));
+    }
+
+    @Override
+    public Map<String, Set<ValidationError>> validateEvents(String events) throws ValidationException {
+        Map<String, Set<ValidationError>> errorsPerEventType = new HashMap<>();
+        JsonNode eventsNodes = parseData(events);
+        eventsNodes.forEach(event -> {
+            String eventType = event.get("eventType").asText();
```
jkevan, line 123: Could be interesting to make the eventType part of the ValidationException to make it a bit more generic.
This would allow to avoid parsing the event type here, as it is supposed to be done in the **extractEventType** function.

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
```diff
@@ -111,12 +112,52 @@ public boolean isEventValid(String event) {
...
+        return errorsPerEventType;
+    }
+
+    private Set<ValidationError> buildCustomErrorMessage() {
+        ValidationMessage.Builder builder = new ValidationMessage.Builder();
+        builder.customMessage("No Schema found for this event type").format(new MessageFormat("Not used pattern. Message format is required"));
+        ValidationError error = new ValidationError(builder.build());
+        Set<ValidationError> errors = new HashSet<>();
+        errors.add(error);
+        return errors;
```
jkevan, line 152: This part would be simpler if we directly return the String validation error in the response instead of a thirdparty object from the external lib.

jkevan (changes requested):

jkevan (approved):

jkevan (approved):

PR #636, opened 2023-07-13
UNOMI-784: use tasks to perform update_by_query and delete_by_query

jkevan:
In order to avoid SocketTimeoutError on long requests between Unomi and ES.
We will delegate to ES the work to do by using tasks system.


persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -1158,6 +1157,45 @@ protected Boolean execute(Object... args) throws Exception {
         }
     }
 
+    private void startTaskTracker(AbstractBulkByScrollRequest request, TaskSubmissionResponse response) {
```
sergehuber, line 1160: May I suggest to rename this method to waitForTask ? With the name start I was expecting to see another system to wait for the ending or for polling for completion.

sergehuber (commented):
Hello Kevan, 
Thank you for the contribution. I have added some comments to the review. 
Best regards,
  Serge...

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
jkevan: Thx, it's renamed.
At first it was a separate function that's why.

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
```diff
@@ -1158,6 +1167,51 @@ protected Boolean execute(Object... args) throws Exception {
...
+                TaskId taskId = new TaskId(response.getTask());
+                while (true){
+                    Optional<GetTaskResponse> getTaskResponseOptional = client.tasks().get(new GetTaskRequest(taskId.getNodeId(), taskId.getId()), RequestOptions.DEFAULT);
+                    if (getTaskResponseOptional.isPresent()) {
+                        GetTaskResponse getTaskResponse = getTaskResponseOptional.get();
+                        if (getTaskResponse.isCompleted()) {
+                            long millis = getTaskResponse.getTaskInfo().getRunningTimeNanos() / 1_000_000;
+                            long seconds = millis / 1000;
+
+                            logger.info("Waiting task [{}]: Finished in {} {}", taskId,
```
dgriffon, line 1188: This log should be in debug as we can have that information using the task ID by querying ES.

We can also add a debug log at the start of the while(true) clause to get the current execution in case of a trouble for one executing (not finished) task.

dgriffon (approved):

persistence-elasticsearch/core/src/main/java/org/apache/unomi/persistence/elasticsearch/ElasticSearchPersistenceServiceImpl.java
jkevan: thx for the review, I will move this log to debug.
But I'm not sure for adding a log each execution, a soon as we have the info on the task started we can track it in ES side, and logging every 1sec will make this difficult to read.

PR #640, opened 2023-08-10
UNOMI-787 : migrate view event page path property

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-787

### DIFF
--- a/tools/shell-commands/src/main/resources/requestBody/2.4.0/base_update_by_query_request.json
+++ b/tools/shell-commands/src/main/resources/requestBody/2.4.0/base_update_by_query_request.json
@@ -1,0 +1,22 @@
+{
+  "script": {
+    "source": "#painless",
+    "lang": "painless"
+  },
+  "query": {
+    "bool": {
+      "must": [
+        {
+          "match": {
+            "itemType": "event"
+          }
+        },
+        {
+          "match": {
+            "eventType": "view"
+          }
+        }
+      ]
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


tools/shell-commands/src/main/resources/requestBody/2.4.0/base_update_by_query_request.json
```diff
@@ -0,0 +1,22 @@
+{
+  "script": {
+    "source": "#painless",
+    "lang": "painless"
+  },
+  "query": {
```
jkevan, line 6: Could we identify the items to update directly from the query ?
meaning the check are still good in the painless, but if the query is able to return less items to be updated by the painless it will be probably faster to execute.

being able to query only document with: `target.properties.path` that exists

tools/shell-commands/src/main/resources/requestBody/2.4.0/view_event_page_path_migrate.painless
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+if (ctx._source.target != null && ctx._source.target.properties != null && ctx._source.target.properties.path != null) {
+    if (ctx._source.target.properties.pageInfo != null) {
```
jkevan, line 19: Should we not create the pageInfo in case it's null ?
currently if it's null the property: `target.properties.path` will stay.

jkevan (changes requested):

jkevan (approved):

PR #666, opened 2024-06-03
UNOMI-816 : ensure empty scope are filled properly at migration time

dgriffon:
Improve scope migration for login event. 
Add Integration test to cover the migration

### DIFF
--- a/itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
+++ b/itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
@@ -19,7 +19,6 @@ package org.apache.unomi.itests.migration;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.unomi.api.*;
-import org.apache.unomi.api.conditions.Condition;
 import org.apache.unomi.itests.BaseIT;
 import org.apache.unomi.persistence.spi.aggregate.TermsAggregate;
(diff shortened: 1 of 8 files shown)


itests/src/test/java/org/apache/unomi/itests/migration/Migrate16xTo220IT.java
```diff
@@ -155,23 +156,23 @@ private void checkForMappingUpdates() throws IOException {
     /**
      * Data set contains a form event (id: 7b55b4fd-5ff0-4a85-9dc4-ffde322a1de6) with this data:
      * {
-     *   "properties": {
```
jsinovassin, line 158: It would be more readable if we keep the indent in this comment.

tools/shell-commands/src/main/resources/requestBody/2.0.0/event_migrate.painless
```diff
@@ -18,7 +18,7 @@
 /* Handle login events */
 if ('login' == ctx._source.eventType) {
     /* Look for empty scope */
-    if (ctx._source.scope == '') {
+    if (ctx._source.scope == '' || ctx._source.scope == null) {
```
jsinovassin, line 21: It would work for user who are still in a using Unomi < 2.0.0 but for users who are already in version 2.0.0, this migration script will not be executed.
It would be better to have another script which will be executed in 2.5.0.

dgriffon: Indeed, it has been moved to a 2.5.0 migration step

jsinovassin (approved):

PR #668, opened 2024-06-07
upgrade cxf

jsinovassin:


jsinovassin:
This PR is done to update CXF, for the moment some dependencies has been updated but karaf is stuck at startup. There is no logs to inform what is explain why karaf is stuck

Fgerthoffert:
Discussion during Unomi project meeting:
 - It might be better to first update karaf, then update CXF, versus the other way around.
 - There might actually be simpler JAX-RS alternatives allowing us to remove CXF altogether.

No concrete action on the short term (other than looking into upgrading Karaf).

PR #687, opened 2024-08-22
UNOMI-846: improve log for event validation

jsinovassin:
https://issues.apache.org/jira/browse/UNOMI-846

### DIFF
--- a/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
+++ b/extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
@@ -85,9 +85,9 @@ public class SchemaServiceImpl implements SchemaService {
         try {
             JsonNode jsonNode = parseData(data);
             JsonSchema jsonSchema = getJsonSchema(schemaId);
-            return validate(jsonNode, jsonSchema).size() == 0;
+            return validate(jsonNode, jsonSchema).isEmpty();
         } catch (ValidationException e) {
-            LOGGER.debug("{}", e.getMessage(), e);
+            LOGGER.warn("{}", e.getMessage(), e);
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


jkevan (changes requested):
Despite the fact that I think changing this log levels directy in the code is a bad idea, mostly because of:
- security
- amount of log printed by default
- the fact that normally log level is configurable (should be configurable) outside the code.

There is inconsistent logs, I wonder why sometimes for a ValidationException we now sometimes log it as:
```
 LOGGER.warn("An event was not valid - switch to DEBUG log level for more information.");
            LOGGER.debug("{}", e.getMessage(), e);
```
and sometimes like this:
```
 LOGGER.warn("{}", e.getMessage(), e);
```

Also I am not really sure about the wording in the message: "An event was not valid" in case of ValidationException, to me it's more a crash/error before/during the validation. Which is different than schema validation results.
I would rename the log to: "An error occurred during event validation", to make that difference, if the goal of thoses logs are to be clear for developers to have useful/understandable informations, they should be clear and consistent.

extensions/json-schema/services/src/main/java/org/apache/unomi/schema/impl/SchemaServiceImpl.java
```diff
@@ -226,9 +228,9 @@ private Set<ValidationError> validate(JsonNode jsonNode, JsonSchema jsonSchema)
             Set<ValidationMessage> validationMessages = jsonSchema.validate(jsonNode);
 
             if (LOGGER.isDebugEnabled() && !validationMessages.isEmpty()) {
-                LOGGER.debug("Schema validation found {} errors while validating against schema: {}", validationMessages.size(), jsonSchema.getCurrentUri());
```
jkevan, line 229: `LOGGER.isDebugEnabled() `should be removed

jkevan (approved):

bpapez (approved):

PR #697, opened 2024-10-04
[UNOMI-853] Adapt migration job to use asynchronous mode avoiding timeout and connection lost

jayblanc:
[UNOMI-853](https://issues.apache.org/jira/browse/UNOMI-853) **Adapt migration job to use asynchronous mode avoiding timeout and connection lost**

Reindex step during migration has been moved to a async task and check against task completion is done until task finish.  
Update a link to karaf in the doc that was broken.

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
@@ -286,13 +286,17 @@ public class MigrationUtils {
     public static void waitForTaskToFinish(CloseableHttpClient httpClient, String esAddress, String taskId, MigrationContext migrationContext) throws IOException {
         while (true) {
             final JSONObject status = new JSONObject(
-                    HttpUtils.executeGetRequest(httpClient, esAddress + "/_tasks/" + taskId + "wait_for_completion=true&timeout=10s",
+                    HttpUtils.executeGetRequest(httpClient, esAddress + "/_tasks/" + taskId + "?wait_for_completion=true&timeout=10s",
                             null));
-            if (status.get("completed").equals("true")) {
+            if (status.has("completed") && status.getBoolean("completed")) {
                 migrationContext.printMessage("Task is completed");
                 break;
             }
-            migrationContext.printMessage("Waiting for Task to complete, current status is " + status.get("status"));
+            if (status.has("error")) {
+                final JSONObject error = status.getJSONObject("error");
+                throw new IOException("Task error: " + error.getString("type") + " - " + error.getString("reason"));
+            }
+            migrationContext.printMessage("Waiting for Task " + taskId + " to complete");
         }
     }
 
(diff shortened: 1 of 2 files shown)


tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
```diff
@@ -216,7 +216,9 @@ public static void reIndex(CloseableHttpClient httpClient, BundleContext bundleC
             // Recreate the original index with new mappings
             HttpUtils.executePutRequest(httpClient, esAddress + "/" + indexName, newIndexSettings, null);
             // Reindex data from clone
-            HttpUtils.executePostRequest(httpClient, esAddress + "/_reindex", reIndexRequest, null);
+            JSONObject task = new JSONObject(HttpUtils.executePostRequest(httpClient, esAddress + "/_reindex?wait_for_completion=false", reIndexRequest, null));
+            //Wait for the reindex task to finish
+            waitForTaskToFinish(httpClient, esAddress, task.getString("task"), migrationContext);
```
jsinovassin, line 221: A reindex query is also called in the method MigrationUtils.moveToIndex, this one might also need to be modified like it's done here

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
```diff
@@ -281,6 +283,23 @@ public static void waitForYellowStatus(CloseableHttpClient httpClient, String es
...
+                            null));
+            if (status.has("completed") && status.getBoolean("completed")) {
+                migrationContext.printMessage("Task is completed");
+                break;
+            }
+            if (status.has("error")) {
+                final JSONObject error = status.getJSONObject("error");
+                throw new IOException("Task error: " + error.getString("type") + " - " + error.getString("reason"));
+            }
+            migrationContext.printMessage("Waiting for Task " + taskId + " to complete");
```
jsinovassin, line 299: Don't we have a risk to have too much logs here. 
According to the time of the reindex, we might have too much log displayed.

jsinovassin (changes requested):

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
jayblanc: Yes I know but migration is not something regular, and knowing that something is running may be useful. What should be the frequency for that log ? 5mn ?

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
jayblanc: Fixed

tools/shell-commands/src/main/java/org/apache/unomi/shell/migration/utils/MigrationUtils.java
jsinovassin: Every 20-30 secondes is a good option I think

jsinovassin (approved):

PR #698, opened 2024-10-04
[UNOMI-854] Add a HealthCheck Endpoint

jayblanc:
[UNOMI-854](https://issues.apache.org/jira/browse/UNOMI-854) **Add a Healthcheck Endpoint**

Add an extension that acts as a healthcheck endpoint. 
The extension uses async call to underlying service to retrieve UP, LIVE, DOWN or ERROR status.  
The health is available in Karaf before unomi starts allowing to check for some config/connection status before starting unimi (like elasticsearch)  
The currently included check are about karaf, elastic search, unomi modules, cluster (cellar) state, unomi persistence.
The Healthcheck has its own configuration and may use authentication based on internal Karaf JAAS modules.  

See included README for more informations.

Note : I tried to use the felix healthcheck https://github.com/apache/felix-dev/blob/master/healthcheck/README.md but I faced wiring issues related to servlet.context package. 


extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/UnomiBundlesHealthCheckProvider.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override public HealthCheckResponse execute() {
+        LOGGER.debug("Health check unomi bundles");
+        HealthCheckResponse.Builder builder = new HealthCheckResponse.Builder();
+        builder.name(NAME).down();
+        try {
+            if (service != null) {
+                builder.up();
+                if (service.isStartupComplete()) {
```
jsinovassin, line 65: **(Not required for this ticket)** In the case when isStartupComplete returns false without failing, it would be nice to have the list of not started module

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/UnomiBundlesHealthCheckProvider.java
```diff
@@ -0,0 +1,74 @@
...
+                    builder.live();
+                }
+            }
+        } catch (Exception e) {
+            builder.error().withData("error", e.getMessage());
```
jsinovassin, line 70: A log here might help as well

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/servlet/HealthCheckHttpContext.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public boolean handleSecurity(HttpServletRequest req, HttpServletResponse res) throws IOException {
+        if (req.getHeader("Authorization") == null) {
+            LOGGER.info("No Authorization header found, sending 401");
+            res.addHeader("WWW-Authenticate", "Basic realm=\"" + realm + "\"");
+            res.sendError(HttpServletResponse.SC_UNAUTHORIZED);
+            return false;
+        }
+        if (authenticated(req)) {
+            LOGGER.info("User authenticated, allowing access");
```
jsinovassin, line 58: This log might generated too much log line according to the number of time the healthcheck is called

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/servlet/HealthCheckHttpContext.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    protected boolean authenticated(HttpServletRequest request) {
+        request.setAttribute(AUTHENTICATION_TYPE, HttpServletRequest.BASIC_AUTH);
+
+        String authzHeader = request.getHeader("Authorization");
+        String usernameAndPassword = new String(Base64.getDecoder().decode(authzHeader.substring(6).getBytes()));
+        String[] parts = usernameAndPassword.split(":");
+
+        LOGGER.info("Authenticating user {}", parts[0]);
```
jsinovassin, line 74: The info logs in this method would be better in debug to avoid too much log lines displayed

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/servlet/HealthCheckServlet.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HealthCheckServlet.class.getName());
+
+    private final HealthCheckService service;
+    private final ObjectMapper mapper;
+
+    public HealthCheckServlet(HealthCheckService service) {
+        LOGGER.info("Building healthcheck servlet...");
+        this.service = service;
+        mapper = new ObjectMapper();
```
jsinovassin, line 48: There is the class CustomObjectMapper which provides the object mapper which is used in UNOMI, can we use it instead here?
https://github.com/apache/unomi/blob/a44710fac7fb6f6bc69f24e65e309bc7b5c0871b/persistence-spi/src/main/java/org/apache/unomi/persistence/spi/CustomObjectMapper.java#L46

manual/src/main/asciidoc/configuration.adoc
```diff
@@ -849,3 +849,79 @@ The following permissions are required by Unomi:
...
+  },
+  {
+    "name":"elasticsearch",
+    "status":"LIVE",
+    "collectingTime":6
+  },
+  {
+    "name":"persistence",
+    "status":"DOWN",
+    "collectingTime":0
```
jsinovassin, line 905: Same message than in the README (for the error informations)

jsinovassin (changes requested):

extensions/healthcheck/README.md
```diff
@@ -0,0 +1,75 @@
...
+  },
+  {
+    "name":"elasticsearch",
+    "status":"LIVE",
+    "collectingTime":6
+  },
+  {
+    "name":"persistence",
+    "status":"DOWN",
+    "collectingTime":0
```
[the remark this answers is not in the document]
jayblanc: I didn't succeed in generating errors during test nor during execution...

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/servlet/HealthCheckServlet.java
jayblanc: That mapper has very specific features configured so I'm not sure it fits with the healthcheck
More than that, it is not available at karaf startup without unomi started and adding that kind of dependency is not  good choice for me.

manual/src/main/asciidoc/configuration.adoc
jayblanc: It is hard to generates relevant errors so I don't have sample response for that.

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/PersistenceHealthCheckProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override public HealthCheckResponse execute() {
+        LOGGER.debug("Health check persistence");
+        HealthCheckResponse.Builder builder = new HealthCheckResponse.Builder();
+        builder.name(NAME).down();
+        try {
+            if (service != null) {
+                builder.up();
+                //TODO replace by the expected persistence version when migrations steps will be stored in the persistence service
+                if (!service.query("target", "profiles", null, PropertyType.class).isEmpty()) {
```
jkevan, line 68: I would suggest to use queryCount if it's to calculate if there is profiles.
But it's strange that this check will return down on a new unomi instance without any profiles existing ?
We could persists something, check it exists, cleanup ?
currently returning all profiles seem's a bit of an heavy operation to know if persistence is up and running.

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/PersistenceHealthCheckProvider.java
jkevan, line 1: Should we move dedicated implementation of health checkers to there dedicated bundles ?
for example making the ElasticSearchPersistenceServiceImpl implementing HealthCheckProvider ?
Since the architecture with the OSGI reference list of implements tend to encourage implementations in other bundles. (just a question, not saying we should do it.)

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/PersistenceHealthCheckProvider.java
jayblanc: This check is only to validate the Peristence layer of unomi and the.
This is a Service of unomi so when not started, that service is not available. It is not about elasticsearch or content presence.
Thus it is not loading profiles but only PropertyType related to profiles. It is the smallest request I found to check that the persistence layer (aka ORM) is configured and works as expected.

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/PersistenceHealthCheckProvider.java
jayblanc: As the health check should run before any startup of unomi, moving checks to dedicated bundles will not allows to use them before any startup time.

extensions/healthcheck/src/main/java/org/apache/unomi/healthcheck/provider/PersistenceHealthCheckProvider.java
jayblanc: Also I had in mind to also check the data model version (migration OK) in ES by integrating something like elasticsearch-evolution. Thus the EalsticSearch healthcheck is about ES connectivity and cluster health (before unomi start) and the Persistence is more about a Check about ORM layer availability. It should include underlying persistence data model version alignement check with the Persistence Layer expected one, like hibernate does at startup.

PR #704, opened 2024-10-23
UNOMI-861: use asynchronous delete_by_query in migration scripts to a…

jsinovassin:
…void timeout
https://issues.apache.org/jira/browse/UNOMI-861
Following the update done in https://github.com/apache/unomi/pull/697

In the migration steps, the delete_by_query and update_by_query has been update to execute an asynchronous call in order to avoid to reach a timeout if the request takes too long.

A try catch has also been added in the the Activate method of UnomiManagementServiceImpl because when there was a fail, the Activate method was called again

### DIFF
--- a/tools/shell-commands/src/main/java/org/apache/unomi/shell/services/internal/UnomiManagementServiceImpl.java
+++ b/tools/shell-commands/src/main/java/org/apache/unomi/shell/services/internal/UnomiManagementServiceImpl.java
@@ -27,6 +27,8 @@ import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


tools/shell-commands/src/main/java/org/apache/unomi/shell/services/internal/UnomiManagementServiceImpl.java
```diff
@@ -38,26 +40,32 @@
...
 
-        if (StringUtils.isNotBlank(bundleContext.getProperty("unomi.autoStart")) &&
-                bundleContext.getProperty("unomi.autoStart").equals("true")) {
-            startUnomi();
+            if (StringUtils.isNotBlank(bundleContext.getProperty("unomi.autoStart")) &&
+                    bundleContext.getProperty("unomi.autoStart").equals("true")) {
+                startUnomi();
+            }
+        } catch (Exception e) {
+            LOGGER.error("Error while starting Unomi", e);
```
dgriffon, line 68: This seems to catch more than a starting Unomi exception.

dgriffon (approved):

PR #728, opened 2025-09-02
UNOMI-906: Fix javadoc tags and new lines.

jayblanc:
This PR fixes some issues that prevent javadoc generation.


Fgerthoffert (approved):

sergehuber:
Actually I have fixed that in the 2.7.x release branch and wanted to backport it to the main branch after completing the release. I solved this differently by closing the <p> tag before the <ul> because the list tags are not allowed inside the <p> tags. But it seems to work fine otherwise when keeping the </p> tag. You can find my commit here, you will notice it requires a lot less changes : 
https://github.com/apache/unomi/commit/d06b31bf324ee71686ce15bbc69d0fa601c1cf6b

PR #747, opened 2026-01-08
chore: configure github settings to only enable squash commits

Fgerthoffert:
During the monthly project team meeting (participants: @sergehuber @jayblanc @jsinovassin @Fgerthoffert ) we discussed only enabling squash commits in GitHub UI to prevent the wrong button from being clicked when merging a PR.


sergehuber (approved):
Hi François, 

Thanks for the PR. Let's hope this works :) 

Regards,
  Serge...

PR #748, opened 2026-01-08
[UNOMI-924] Add healthcheck activated by default

jayblanc:
## Description 

The goal is to activate the Healthcheck by default


sergehuber (approved):
Hi Jérome, 

Thanks for the change, for me it's good.

Best regards,
  Serge...

PR #757, opened 2026-05-14
UNOMI-139/880/878/884: Multi-tenant platform core

sergehuber:
### Stacked PR (merge order)

| Role | Branch |
|------|--------|
| **Base (merge into)** | `UNOMI-888-import-export-javadoc` |
| **Head (this PR)** | `UNOMI-139-UNOMI-880-multitenancy` |

This PR is **stacked**: it should merge **into `UNOMI-888-import-export-javadoc` first**; after that PR is on `master` (or the next stack parent), this slice reduces to the multi-tenant / cache / scheduler / migration / V2-compat delta relative to that line.

**Merge-base with `origin/master`** (when last checked): `c85fb8856` — same tip as `origin/master` at fork time for the underlying stack; the branch contains prior stacked commits (UNOMI-920 → UNOMI-879 → UNOMI-888) before this work.

---

### Why a single PR
These changes share the same execution model (tenant context, unified cache, scheduler, and persistence) and are wired through the same services, blueprints, and integration tests. Splitting them into separate commits would produce artificial intermediate states that neither compile nor pass ITs, and would force heavy merge or rebase work across the same files repeatedly. They are therefore delivered together as one coherent “3.1 platform slice” for review, even though the work maps to several JIRA tickets.

This change set backports and wires the Unomi 3.1 multi-tenant platform work into the integration branch: tenant-aware execution, centralized caching, the new scheduler service, V3 migration assets, and optional V2 REST compatibility. Tracing and full type-validation services from unomi-3-dev are not included; call sites use ParserHelper / defensive handling where those stacks were removed for a stacked follow-up branch.

### UNOMI-139 Multi-tenancy
- Introduce tenant model and APIs (Tenant, TenantService, API keys, security/audit interfaces, quotas, lifecycle hooks) and propagate tenantId on Item and related types.
- Add ExecutionContextManager and thread-local tenant context so services, persistence, and REST layers resolve the active tenant consistently.
- Adjust persistence (Elasticsearch/OpenSearch) and services to scope reads/writes and inheritance (e.g. system tenant vs tenant overrides).

### UNOMI-880 Unified multi-tenant caching
- Add MultiTypeCacheService / CacheableTypeConfig and AbstractMultiTypeCachingService for shared cache lifecycle, bundle predefined items, periodic refresh, and statistics.
- Migrate DefinitionsServiceImpl and SegmentServiceImpl (and related paths) onto the unified cache; remove duplicated ad-hoc cache listeners where superseded.
[...]

PR #760, opened 2026-05-18
UNOMI-139 UNOMI-880 UNOMI-878 UNOMI-884: Multi-tenant platform core

sergehuber:
### Stacked PR (merge order)

| Role | Branch |
|------|--------|
| **Base (merge into)** | `UNOMI-888-import-export-javadoc` |
| **Head (this PR)** | `UNOMI-139-platform` |

This PR is **stacked**: merge **into `UNOMI-888-import-export-javadoc` first** ([PR #756](https://github.com/apache/unomi/pull/756), UNOMI-888), after that line is merged or rebased. Downstream PRs (UNOMI-904 V2 compat, dev shell, additional ITs, manual) will stack on `UNOMI-139-platform`.

For more information (each PR targets the branch below until the bottom merges): https://github.github.com/gh-stack/introduction/overview/

---

**JIRA**

- UNOMI-139: https://issues.apache.org/jira/browse/UNOMI-139
- UNOMI-880: https://issues.apache.org/jira/browse/UNOMI-880
- UNOMI-878: https://issues.apache.org/jira/browse/UNOMI-878
- UNOMI-884: https://issues.apache.org/jira/browse/UNOMI-884

### Summary

Delivers the Unomi 3.1 **platform core** on top of [UNOMI-888](https://github.com/apache/unomi/pull/756): multi-tenant execution, unified caching, cluster-aware scheduling, and V3 migration assets. Services, persistence, GraphQL, and REST are wired for **V3 tenant resolution** (API keys / roles).

Supersedes the platform scope of [PR #757](https://github.com/apache/unomi/pull/757) (closed; monolith branch `UNOMI-139-UNOMI-880-multitenancy`). **UNOMI-904** (V2 compatibility), dev shell commands, additional IT classes, and manual updates will follow in stacked PRs on this line.

### UNOMI-139 Multi-tenancy

- Tenant model and APIs (`Tenant`, `TenantService`, API keys, security/audit interfaces, quotas, lifecycle hooks) and `tenantId` on core types.
- `ExecutionContextManager` and thread-local tenant context so services, persistence, and REST resolve the active tenant consistently.
- Tenant-scoped persistence and definitions (Elasticsearch/OpenSearch), including system-tenant vs tenant-specific overrides.

### UNOMI-880 Unified multi-tenant caching

- `MultiTypeCacheService` / `CacheableTypeConfig` and `AbstractMultiTypeCachingService` for shared cache lifecycle, predefined item bundling, periodic refresh, and statistics.
- Migrate `DefinitionsServiceImpl` and `SegmentServiceImpl` (and related paths) onto the unified cache; remove duplicated ad-hoc cache listeners where superseded.

### UNOMI-878 Cluster-aware task scheduling

- Replace legacy scheduler usage with `SchedulerService` (scheduled tasks, task executors, persistence-backed vs in-memory tasks, cluster coordination).
- Integrate periodic work (cache refresh, cluster heartbeat, router scheduling, rule refresh paths) with the new scheduler.
- Update dependents (e.g. Groovy actions service) to schedule work through the new API.

### UNOMI-884 Migration to V3

- Migration Groovy scripts and request bodies for **3.1.0** (tenant document IDs, system item ID fixes, tenant initialization, legacy queryBuilder updates).
- Extended `MigrationUtils` and `MigrationUtilsTest` for the new steps aligned with multi-tenancy and index updates.

### Tests

- Platform integration tests updated (`BaseIT`, migration ITs, `ScopeIT`, and related suites).
- Full integration test run verified locally with OpenSearch (`./build.sh --integration-tests --use-opensearch`).

### Licence


services-common/src/main/java/org/apache/unomi/services/common/security/KarafSecurityService.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    @Override
+    public boolean hasSystemAccess() {
+        return hasRole(UnomiRoles.ADMINISTRATOR) || hasRole(UnomiRoles.TENANT_ADMINISTRATOR);
+    }
+
+    @Override
+    public boolean hasTenantAccess(String tenantId) {
+        if (hasRole(UnomiRoles.TENANT_ADMINISTRATOR)) {
```
sergehuber: **[C1 CRITICAL] Tenant isolation bypass — `tenantId` parameter is ignored**

Every API-key-authenticated tenant receives `TENANT_ADMINISTRATOR` (see `createSubject()` line 323). This check then returns `true` for **any** `tenantId` when that role is held — Tenant A’s private key grants full read/write access to Tenant B’s data.

Fix:

```java
@Override
public boolean hasTenantAccess(String tenantId) {
    if (hasRole(UnomiRoles.ADMINISTRATOR)) {
        return true; // system admin bypasses tenant boundary
    }
    String subjectTenantId = getCurrentSubjectTenantId();
    return tenantId != null && tenantId.equals(subjectTenantId);
}
```

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/servlet/auth/GraphQLServletSecurityValidator.java
```diff
@@ -133,14 +180,31 @@ private boolean isAuthenticatedUser(HttpServletRequest req) {
...
+                // Check for tenant ID header
+                String tenantId = req.getHeader(UNOMI_TENANT_ID_HEADER);
+                if (tenantId != null && !tenantId.trim().isEmpty()) {
+                    // Validate tenant exists
+                    Tenant tenant = tenantService.getTenant(tenantId);
+                    if (tenant != null) {
+                        executionContextManager.setCurrentContext(executionContextManager.createContext(tenantId));
+                    } else {
+                        LOG.warn("Invalid tenant ID provided in header: {}", tenantId);
+                        executionContextManager.setCurrentContext(ExecutionContext.systemContext());
```
sergehuber: **[C2 CRITICAL] JAAS user with unknown tenant ID receives full system privileges**

When JAAS authentication succeeds but `X-Unomi-Tenant-Id` names a tenant that does not exist, this line grants `systemContext()` — which sets `isSystem = true` and bypasses every `validateAccess()` check. Any Karaf user can exploit this by passing a random non-existent tenant ID.

Safe fix: drop the context override and let the request continue under the JAAS user’s own subject-derived context.

```suggestion
                        // Unknown tenant — do not grant system privileges; leave context derived from JAAS login
```

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-10-tenantInitialization.groovy
```diff
@@ -0,0 +1,88 @@
...
+                  "revoked" : false
+                },
+                {
+                  "itemId" : "3c595ea8-000e-4d0b-a329-0d259cc4d176",
+                  "itemType" : "apiKey",
+                  "createdBy": "system-migration-3.1.0",
+                  "lastModifiedBy": "system-migration-3.1.0",
+                  "creationDate" : "${isoDate}",
+                  "lastModificationDate" : "${isoDate}",
+                  "key" : "503BAABB3A14AEB4B50ACF3C82982FBABECDBAEA83879CA8AECA016A6A9EEA85",
```
sergehuber: **[C3 CRITICAL] Private API key credential hardcoded in source — committed to git history permanently**

Both the public key (`C606D77D…` line 62) and this private key (`503BAABB…`) will bootstrap the system tenant on every migration run. Anyone with repository access has working credentials. Private keys must never be committed to source.

Replace both key values with keys generated at migration time:

```groovy
import java.security.SecureRandom
SecureRandom rng = SecureRandom.getInstanceStrong()
byte[] pubBytes  = new byte[32]; rng.nextBytes(pubBytes)
byte[] privBytes = new byte[32]; rng.nextBytes(privBytes)
String generatedPublicKey  = pubBytes.collect  { String.format('%02X', it) }.join()
String generatedPrivateKey = privBytes.collect { String.format('%02X', it) }.join()
// use generatedPublicKey / generatedPrivateKey in the JSON body
```

The fixed `itemId` UUIDs on lines 56 and 67 compound the issue — an attacker who knows the key values also knows the exact Elasticsearch document IDs. Consider generating those at runtime too.

api/src/main/java/org/apache/unomi/api/tenants/ApiKey.java
```diff
@@ -0,0 +1,204 @@
...
+        /**
+         * Private API key for protected endpoints including login and updateProperties
+         */
+        PRIVATE
+    }
+
+    /**
+     * The API key value.
+     */
+    private String key;
```
sergehuber, line 52: **[C4 CRITICAL] Raw API key value serialised to Elasticsearch and exposed by the REST layer**

This field has no `@JsonIgnore`. Since `Tenant extends Item` and is persisted to Elasticsearch, the entire `List<ApiKey>` — including raw `key` values — is stored in plaintext and returned by `GET /tenants/{id}` to any caller with the `ADMINISTRATOR` role. The `@XmlTransient` annotations on the derived getters suppress JAXB only, not Jackson.

Private keys are authentication credentials and should be stored as a secure hash, returned to the caller only once at creation time.

Minimal immediate fix (also add `import com.fasterxml.jackson.annotation.JsonIgnore;` to the class imports):

```suggestion
    @com.fasterxml.jackson.annotation.JsonIgnore
    private String key;
```

sergehuber (commented):
## Code Review: UNOMI-139 Multi-tenant platform core

**Scope:** 292 files, +30 486 / −4 093 lines (multi-tenancy, unified caching, cluster scheduling, V3 migration).

| # | Severity | Finding |
|---|----------|---------|
| C1 | **CRITICAL** | `hasTenantAccess()` ignores `tenantId` — any tenant key accesses all tenants |
| C2 | **CRITICAL** | Unknown `X-Unomi-Tenant-Id` header grants full system context to JAAS users |
| C3 | **CRITICAL** | Private API key committed in plaintext to migration script (permanently in git history) |
| C4 | **CRITICAL** | `ApiKey.key` serialised to ES and returned by REST with no `@JsonIgnore` |
| H1 | **HIGH** | `executeAsSystem()` finally block suppresses original exception on dual failure |
| H2 | **HIGH** | Migration step key uses `itemType` — multiple rollover indices silently skipped |
| I1 | **IMPORTANT** | `SecurityService.SYSTEM_TENANT = "SYSTEM_TENANT"` conflicts with all other constants (`"system"`) |
| I2 | **IMPORTANT** | `isPublicOperation()` throws `IndexOutOfBoundsException` on empty/fragment-only documents |
| I3 | **IMPORTANT** | `migrate-3.1.0-00-tenantDocumentIds.groovy` conflicts with existing `00-fixProfileNbOfVisits` on master |
| I4 | **IMPORTANT** | `ScheduledTask implements Serializable` without `serialVersionUID` |
| I5 | **IMPORTANT** | `getCurrentRoles()` is dead code calling `AccessController` removed in Java 21 |
| I6 | **IMPORTANT** | `isOperatingOnSystemTenant()` stub always returns `false` |
| A1 | Advisory | `@author` tag in `SchedulerServiceImpl` (ASF policy) |

All findings ≥ 80 % confidence. Inline suggestions attached. C1–C4 should be resolved before merge.

services-common/src/main/java/org/apache/unomi/services/common/security/ExecutionContextManagerImpl.java
```diff
@@ -0,0 +1,191 @@
...
+                if (previousContext != null) {
+                    currentContext.set(previousContext);
+                } else {
+                    currentContext.remove();
+                }
+                securityService.setCurrentSubject(previousSubject);
+            } catch (Exception e) {
+                LOGGER.error("Error restoring previous context: {}", e.getMessage(), e);
+                // Still throw the error to ensure it's not silently ignored
+                throw new SecurityException("Failed to restore security context", e);
```
sergehuber: **[H1 HIGH] Original operation exception suppressed when context restore also fails**

If `operation.get()` throws `E1` and `setCurrentSubject()` then throws `E2` in the `finally` catch, this `throw new SecurityException(…)` replaces `E1`. The caller receives a generic restore-failure message and the actual root cause is silently lost (not even added via `Throwable.addSuppressed()`).

Remove the re-throw — the error is already logged, and rethrowing here swallows `E1` every time both the operation and the restore fail together. Also update the stale comment on line 109.

```suggestion
                // Context restore failed — already logged above. Do not rethrow to avoid suppressing the original operation exception.
```

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-00-tenantDocumentIds.groovy
```diff
@@ -0,0 +1,179 @@
...
+        // Build index settings
+        String newIndexSettings
+        if (indexConfig.useRollover) {
+            newIndexSettings = MigrationUtils.buildIndexCreationRequestWithRollover(baseSettings, mapping, context, rolloverPolicyName, indexConfig.alias(indexPrefix))
+        } else {
+            newIndexSettings = MigrationUtils.buildIndexCreationRequest(baseSettings, mapping, context, false)
+        }
+
+        // Execute reindex
+        MigrationUtils.reIndex(context.getHttpClient(), bundleContext, esAddress, indexName, newIndexSettings, updateScript, params, context, "3.1.0-${itemType}-update")
```
sergehuber: **[H2 HIGH] Non-idempotent step key — multiple rollover indices of the same type silently skipped**

For rollover indices (`session`, `event`) there are typically several physical backing indices that all resolve to the same `itemType`. `performMigrationStep` marks a step `COMPLETED` after the first index of that type is processed; every subsequent same-type index is then silently skipped on the current run and on any retry. Data in those indices is never migrated.

Use the unique physical `indexName` instead of `itemType` as the step key:

```suggestion
        MigrationUtils.reIndex(context.getHttpClient(), bundleContext, esAddress, indexName, newIndexSettings, updateScript, params, context, "3.1.0-${indexName}-update")
```

api/src/main/java/org/apache/unomi/api/security/SecurityService.java
```diff
@@ -0,0 +1,234 @@
...
+ * - Subject management (authentication and authorization)
+ * - Role-based access control (RBAC)
+ * - Tenant isolation and access control
+ * - Operation validation
+ * - System and privileged operations
+ * - Encryption key management
+ */
+public interface SecurityService {
+    /** The system tenant identifier used for system-wide operations */
+    String SYSTEM_TENANT = "SYSTEM_TENANT";
```
sergehuber: **[I1 IMPORTANT] `SecurityService.SYSTEM_TENANT` constant has the wrong value**

Every other definition in the codebase uses `"system"`:
- `ExecutionContext.SYSTEM_TENANT = "system"`
- `TenantService.SYSTEM_TENANT = "system"`
- `KarafSecurityService.SYSTEM_TENANT = "system"`

Any caller who writes `tenantId.equals(SecurityService.SYSTEM_TENANT)` will silently **never match** the system tenant, defeating whatever isolation guard they intended.

```suggestion
    String SYSTEM_TENANT = "system";
```

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/servlet/auth/GraphQLServletSecurityValidator.java
```diff
@@ -74,6 +96,10 @@ public boolean validate(String query, String operationName, HttpServletRequest r
     }
 
     private boolean isPublicOperation(String query) {
+        if (query == null) {
+            return false;
+        }
+
         final Document queryDoc = parser.parseDocument(query);
         final Definition<?> def = queryDoc.getDefinitions().get(0);
```
sergehuber, line 107: **[I2 IMPORTANT] `IndexOutOfBoundsException` on empty or fragment-only GraphQL documents**

`getDefinitions().get(0)` throws if the parsed document has no operation definitions (e.g. a fragment-only document, or a body that parses without error but contains no operations). Because `isPublicOperation` runs before the `Authorization` check, this is reachable unauthenticated and returns a 500 instead of a clean 400.

```suggestion
        if (queryDoc.getDefinitions().isEmpty()) {
            return false;
        }
        final Definition<?> def = queryDoc.getDefinitions().get(0);
```

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-01-tenantDocumentIds.groovy
```diff
@@ -0,0 +1,179 @@
+import org.apache.unomi.shell.migration.service.MigrationContext
```
sergehuber, line 1: **[I3 IMPORTANT] Filename conflict — `migrate-3.1.0-00-fixProfileNbOfVisits.groovy` already exists on `master`**

`migrate-3.1.0-00-fixProfileNbOfVisits.groovy` was added by UNOMI-922 (commit `c0f145bb0`) and is already in `master`. Both scripts share step priority `0`. They both execute (the `TreeSet` orders them alphabetically by name when priority is equal, so `fixProfileNbOfVisits` runs first), but:

1. `migrate-3.1.0-05-fixSystemItemIds.groovy` refers to _“the 3.1.0-00 migration script”_ in the singular — now ambiguous.
2. Future operators and developers cannot tell which `00` was intentional.

Please rename this file to `migrate-3.1.0-01-tenantDocumentIds.groovy` (or another unambiguous number that makes the ordering relative to `fixSystemItemIds` explicit).

api/src/main/java/org/apache/unomi/api/tasks/ScheduledTask.java
```diff
@@ -0,0 +1,873 @@
...
+ *   <li>Lock management for cluster coordination</li>
+ *   <li>Execution history and checkpoint data for recovery</li>
+ *   <li>Support for one-shot and periodic execution</li>
+ *   <li>Task dependencies and parallel execution control</li>
+ *   <li>Cluster-wide task distribution</li>
+ * </ul>
+ */
+public class ScheduledTask extends Item implements Serializable {
+
+    public static final String ITEM_TYPE = "scheduledTask";
```
sergehuber, line 44: **[I4 IMPORTANT] `Serializable` class without `serialVersionUID` — rolling-upgrade hazard**

`ScheduledTask` implements `Serializable` and has ~25 fields (`Map`, `Set`, `TimeUnit`, `Date`, inner `TaskStatus` enum). Without a declared `serialVersionUID` the JVM computes one from the class structure. Adding or removing any field will invalidate persisted serialised blobs in a rolling cluster upgrade and throw `InvalidClassException`.

```suggestion
    private static final long serialVersionUID = 1L;

    public static final String ITEM_TYPE = "scheduledTask";
```

services-common/src/main/java/org/apache/unomi/services/common/security/ExecutionContextManagerImpl.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Override
+    public void executeAsTenant(String tenantId, Runnable operation) {
+        executeAsTenant(tenantId, () -> {
+            operation.run();
+            return null;
+        });
+    }
+
+    private Set<String> getCurrentRoles() {
```
sergehuber: **[I5 IMPORTANT] Dead code calling `AccessController.getContext()` — removed in Java 21**

`getCurrentRoles()` calls `AccessController.getContext()` which was deprecated for removal in Java 17 and removed entirely in Java 21. A project-wide search confirms this method has **zero call sites** — it is unreachable dead code.

If ever invoked on a Java 21+ runtime it will throw `NoSuchMethodError`. Please delete the entire method (lines 155–166). If the functionality is ever needed it should be re-implemented using the `securityService` instance already available in this class.

services-common/src/main/java/org/apache/unomi/services/common/security/KarafSecurityService.java
```diff
@@ -0,0 +1,333 @@
...
+            if (!tenantPrincipals.isEmpty()) {
+                return tenantPrincipals.iterator().next().getTenantId();
+            }
+        }
+        return SYSTEM_TENANT;
+    }
+
+    @Override
+    public boolean isOperatingOnSystemTenant() {
+        return false;
```
sergehuber: **[I6 IMPORTANT] `isOperatingOnSystemTenant()` is a non-functional stub — always returns `false`**

This method is part of the `SecurityService` API contract. Any caller that uses it to guard a system-only operation will silently never activate that guard.

```suggestion
        return getCurrentSubjectTenantId().equals(SYSTEM_TENANT);
```

services/src/main/java/org/apache/unomi/services/impl/scheduler/SchedulerServiceImpl.java
```diff
@@ -17,57 +17,2015 @@
...
+ * - Tasks can be configured to run on specific nodes or all nodes
+ * - Lock ownership prevents duplicate execution
+ * - Crash recovery handles node failures
+ *
+ * Pending Operations:
+ * - Operations that require subservices are queued during initialization
+ * - Operations are executed once all required services are available
+ * - Supports different operation types with appropriate handling
+ *
  * @author dgaillard
```
sergehuber: **[A1 Advisory] `@author` tag — violates ASF attribution policy**

The Apache Software Foundation asks projects not to use `@author` Javadoc tags; author attribution belongs in `git log`.

```suggestion
 *
```

services-common/src/main/java/org/apache/unomi/services/common/security/KarafSecurityService.java
sergehuber: Fixed! The method now properly compares the requested tenantId against the subject's own tenant ID, so a tenant principal can only access their own tenant. We kept a blanket true for ADMINISTRATOR role only. We also took the opportunity to fix I6 at the same time — isOperatingOnSystemTenant() was a stub always returning false and now correctly checks against SYSTEM_TENANT.

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/servlet/auth/GraphQLServletSecurityValidator.java
sergehuber: Fixed! Rather than falling back to systemContext() when an unknown tenant ID is supplied in the header, we now leave the JAAS-derived context untouched. This closes the privilege escalation path where an attacker could supply a random header value to gain system access.

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-10-tenantInitialization.groovy
sergehuber: Fixed in commit 9eefd885: rather than hardcoding any key, the migration script now generates fresh cryptographically-random keys at migration time using SecureRandom. The keys are printed to the console once (the only opportunity to record them) and also written to a time-limited file under karaf.data/migration/secrets/ that is automatically cleaned up 24 hours after creation by the next migration run. This way no secret ever touches the git repository, and the operator is clearly prompted to save the keys to a vault. The migration step is idempotent — key generation only runs when the tenant index doesn't yet exist.

api/src/main/java/org/apache/unomi/api/tenants/ApiKey.java
sergehuber: Thanks for flagging this. We've raised [UNOMI-938](https://issues.apache.org/jira/browse/UNOMI-938) to track the fix that will be part of another PR before release of Unomi 3.1.0

services-common/src/main/java/org/apache/unomi/services/common/security/ExecutionContextManagerImpl.java
sergehuber: Good spot. The finally block was doing a context restore and throwing its own exception on failure, which would silently swallow the original exception from the operation itself — very hard to debug. Fixed in commit 9eefd885: we now just log the restore failure and let the original exception propagate naturally. We also took this opportunity to remove the dead getCurrentRoles() method (I5) that was calling AccessController which no longer exists in Java 21.

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-00-tenantDocumentIds.groovy
sergehuber: Fixed in commit 9eefd885: migrate-3.1.0-10-tenantInitialization.groovy was updated to use the correct index-specific step keys, ensuring rollover index types like session and event are no longer silently skipped during migration.

api/src/main/java/org/apache/unomi/api/security/SecurityService.java
sergehuber: Fixed in commit 9eefd885: corrected to "system". This was causing subtle context resolution failures across services that used TenantService.SYSTEM_TENANT alongside SecurityService.SYSTEM_TENANT — they were comparing different strings for what should be the same concept.

graphql/cxs-impl/src/main/java/org/apache/unomi/graphql/servlet/auth/GraphQLServletSecurityValidator.java
sergehuber: Fixed in commit 9eefd885: GraphQLServletSecurityValidator now guards against an empty definitions list before accessing index 0, returning false for malformed/empty documents rather than throwing.

tools/shell-commands/src/main/resources/META-INF/cxs/migration/migrate-3.1.0-01-tenantDocumentIds.groovy
sergehuber: Fixed in commit 9eefd885: renamed from migrate-3.1.0-00-... to migrate-3.1.0-01-... to avoid the numbering collision with the existing master script.

api/src/main/java/org/apache/unomi/api/tasks/ScheduledTask.java
sergehuber: Fixed in commit 9eefd885: added private static final long serialVersionUID = 1L. Without this, any change to the class structure would break deserialization across cluster nodes that haven't all been upgraded simultaneously.

services-common/src/main/java/org/apache/unomi/services/common/security/ExecutionContextManagerImpl.java
sergehuber: Fixed in commit 9eefd885: getCurrentRoles() and its AccessController.getContext() call have been removed. This code was unreachable and would have caused a NoSuchMethodError at runtime on Java 21+.

services-common/src/main/java/org/apache/unomi/services/common/security/KarafSecurityService.java
sergehuber: Fixed in commit 9eefd885 alongside C1: now correctly delegates to getCurrentSubjectTenantId().equals(SYSTEM_TENANT).

services/src/main/java/org/apache/unomi/services/impl/scheduler/SchedulerServiceImpl.java
sergehuber: Fixed in commit 9eefd885: @author tag removed. Per ASF policy, authorship is tracked via git history rather than inline annotations.

PR #853, opened 2026-08-14
UNOMI-977: Restrict router import/export configuration to system administrators

sergehuber:
A router import or export configuration carries a Camel source or destination URI, so writing one
reaches the filesystem and remote endpoints. That is a host-level capability rather than one
confined to a tenant's own data, and it should require the corresponding role.

ImportConfigurationServiceEndPoint and ExportConfigurationServiceEndPoint now require
UnomiRoles.ADMINISTRATOR, on the oneshot and multipart upload paths as well as the CRUD ones.

RouterEndpointRoleSecurityIT exercises all four over HTTP rather than inspecting the annotations. A
role annotation that is present but not wired - an endpoint outside the security filter's scope, say
- would still satisfy a unit test that only reads the annotation.

Jira: https://issues.apache.org/jira/browse/UNOMI-977

### DIFF
--- a/extensions/router/router-karaf-feature/src/main/feature/feature.xml
+++ b/extensions/router/router-karaf-feature/src/main/feature/feature.xml
@@ -20,6 +20,8 @@
         <details>Apache Karaf feature for the Apache Unomi Context Server extension</details>
         <feature>wrap</feature>
         <feature>unomi-services</feature>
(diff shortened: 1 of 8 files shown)


sergehuber:
Rebased onto #852 so the `AllITs` / `CorePersistenceITs` suite entries compose. Merge #852 first; after that this should retarget onto `master`.

PR #861, opened 2026-09-15
UNOMI-980: create the compatibility tenant, and name the mode for what it does

jsinovassin:
[UNOMI-980](https://issues.apache.org/jira/browse/UNOMI-980)

### What this changes

The compatibility mode lets a client written before Unomi 3.1 keep working with no change on its side. It refused every request until an operator created a tenant by hand, which is a step no earlier version asked for:

```
ERROR AuthenticationFilter - V2 compatibility mode: configured default tenant 'default' does not exist
```

Three arms, all in that mode.

**Create the tenant the mode runs on.** The new `TenantService.getOrCreateTenant` is synchronized, and it re-reads the tenant when a concurrent creation wins the race. Several nodes starting together all get the tenant, and only one creation happens.

**Drop the tenant identifier setting.** The mode read the setting in two places, and both sit inside the method that runs only when the mode is on. A client from before 3.1 knows no tenant, so it never names one, and nothing outside the filter read the value. The identifier is now a constant.

Two unreachable branches go with the setting. Each call site tested the value for blankness, while `modified()` already fell back to `default`. The two branches also disagreed: a blank value denied every public request on one path, and switched to the system context on the other.

**Rename the mode.** Unomi 3.0.0 and 3.0.1 carry no tenant either, so a 3.0.1 client meets the same wall as a V2 client. The old name named the wrong thing. `v2.compatibilitymode.enabled` becomes `singletenantcompatibility.enabled`. The 3.1 line is not released, so the key can still change.

### Tests

`V2CompatibilityModeIT` becomes `SingleTenantCompatibilityModeIT`. Its two tests that existed only to exercise the removed setting are gone.

Two tests built a fixture through the services, which puts it in the tenant `BaseIT` works in. They then read the fixture back through the compatibility path, which runs on its own tenant. `testV2ModeBehavior` now reads back the profile its own V2-style request created. `setUp` now creates the test scope in the compatibility tenant as well, and puts the execution context back afterwards.

```
itests SingleTenantCompatibilityModeIT   3 tests, 0 failures
unomi-rest unit tests                    120 tests, 0 failures
```

The full integration suite has not been run.


sergehuber (commented):
Initial review.

`getOrCreateTenant` in compatibility mode is the right fix: a 2.x client cannot name a tenant, so this mode has to create `default` rather than 401 until an operator does it by hand. The create/re-read race handling looks correct.

Renaming `v2.compatibilitymode.enabled` to `singletenantcompatibility.enabled` is fine while 3.1 is unreleased.

Please make the docs match the code on the tenant id. The PR drops `v2.compatibilitymode.defaultTenantId` and hard-codes `default`, but `migrate-3.0-to-3.1.adoc` and `v2-compatibility-mode.adoc` still tell operators to set `v2.compatibilitymode.tenantId`. That setting will not do anything. Same check for leftover `UNOMI_REST_AUTHENTICATION_V2COMPATIBILITYMODE*` environment names versus the new `SINGLETENANTCOMPATIBILITY` form.

CI is green. Happy to merge once the docs match.

sergehuber:
Thank you for this, Jonathan — it is a really helpful piece of 3.1 migration work, and I am grateful you took the time to make compatibility mode actually usable for 2.x / 3.0 clients.

`getOrCreateTenant` is the right idea: a legacy client cannot name a tenant, so this mode should create `default` rather than answering 401 until someone does it by hand. The create / re-read race handling looks solid, and renaming `v2.compatibilitymode.enabled` to `singletenantcompatibility.enabled` is a nice clarity improvement while 3.1 is still unreleased. CI looking green is much appreciated as well.

One small docs follow-up, if you have a moment — happy to help with the wording if useful. The code now hard-codes tenant `default` and drops `v2.compatibilitymode.defaultTenantId`, but `migrate-3.0-to-3.1.adoc` and `v2-compatibility-mode.adoc` still tell operators to set `v2.compatibilitymode.tenantId`. That setting will no longer do anything, so people following the guide could think they had configured a tenant when they had not. The same pass on leftover `UNOMI_REST_AUTHENTICATION_V2COMPATIBILITYMODE*` environment names versus the new `SINGLETENANTCOMPATIBILITY` form would save operators a hunt.

Thank you again for this. Happy to merge as soon as the docs match the code, and please do keep the PRs coming.

PR #864, opened 2026-09-16
UNOMI-982: Drop a condition whose optional parameter the context does not supply

jsinovassin:
Fixes UNOMI-982, at https://issues.apache.org/jira/browse/UNOMI-982

### What this changes

A condition type can declare an optional parameter, and its parent condition then compares a field to `parameter::<name>`. When the caller supplies no value for that name, the comparison states no constraint, so the evaluator has to ignore the comparison. Apache Unomi 3.1 evaluates the comparison instead, so it becomes `field equals null` and matches nothing. A `booleanCondition` with the `and` operator holds that comparison, so the whole condition never matches, and every rule built on the condition type stops firing.

The failure is silent. Unomi logs no warning and no error, the rule stays enabled and valid in persistence, and the REST API returns it unchanged.

### Where the behaviour changed

UNOMI-883 changed three places, at https://github.com/apache/unomi/commit/901b5fcdc62a7732097f7db9305dd9404356f8a1 and https://github.com/apache/unomi/pull/773.

`ConditionContextHelper.parseParameterWithValidationRecursive` keeps the unresolved value in the parameter map. Before that commit, an unresolved parameter reference voided the whole condition.

`ConditionEvaluatorDispatcherImpl.eval` answers `false` for a null contextual condition. Before that commit it answered `true`, which means the caller drops the condition.

`ConditionESQueryBuilderDispatcher.buildFilter` answers a match-none query for a null contextual condition. Before that commit it answered a match-all query. `ConditionOSQueryBuilderDispatcher.buildFilter` has the same shape.

### Mechanism

`getContextualCondition` now gives three answers instead of two:

- the resolved condition,
- `null`, when the context does not supply a parameter reference the condition carries,
- a new `ConditionContextHelper.UNRESOLVABLE` sentinel, when resolution itself failed.

A cyclic reference, a depth overrun and a script that did not run all produce `UNRESOLVABLE`. The callers ignore a `null` condition and refuse an `UNRESOLVABLE` one, so UNOMI-883 keeps its guard on the broken cases.

### Effect

Measured on Apache Unomi 3.1.0-SNAPSHOT, with a module hat declares `pageViewEventCondition`, whose `pagePath` and `language` parameters are optional.

A `view` event carrying URL parameters fired neither the `copyURLParams` rule nor the `incrementPageViewCount` rule:

```
{"nbOfVisits":1,"lastVisit":"...","firstVisit":"...","totalNbOfVisits":1}
```

Both rules fire again with this change:

```
{"nbOfVisits":1,"pageViewCount":{"copyURLParams":1},"lastVisit":"...","firstVisit":"...","totalNbOfVisits":1,"utm_content":["testParamValue"]}
```

### Tests

`mvn test` on `persistence-spi` reports 134 tests and no failure.

Eleven tests in `ConditionContextHelperTest` asserted the behaviour this pull request reverses, and they now assert the two answers apart. `ConditionEvaluatorDispatcherImplTest` gains two tests: an unset parameter drops the condition, and a cyclic reference refuses the condition.

### Notes for the reviewer

The two past event query builders, `PastEventConditionESQueryBuilder` and `PastEventConditionOSQueryBuilder`, still throw on a null contextual condition. They call `getContextualCondition` on the top level event condition, whose own parameter values hold no reference, so this change does not alter their behaviour.

The `List` branch of `parseParameterWithValidationRecursive` still keeps a `null` element for an unresolved reference. Apache Unomi 2.x skipped that element. No test covers the difference, so this pull request leaves the branch alone.


sergehuber:
Thank you for this, Jonathan — this is an important 3.1 regression fix and I am really glad you found it.

Optional `parameter::` values that are absent turning into `field == null` (and then failing the whole `and` condition) would quietly stop rules such as `copyURLParams` and page-view counting. Distinguishing “unset optional parameter” (`UNRESOLVABLE`) from “resolved to null” is a careful restoration of the 2.x “drop this clause” behaviour, without undoing UNOMI-883. Using identity comparison against the sentinel is the right way to read it.

CI is green, including the ITs, which is much appreciated. I am happy to merge this. Thank you for taking the time — contributions like this help a lot, and they are very welcome.

PR #865, opened 2026-09-16
UNOMI-983: Refuse to save an item that carries no itemId

jsinovassin:
Fixes UNOMI-983, at https://issues.apache.org/jira/browse/UNOMI-983

### What this changes

An item states its own identity. Apache Unomi 3.1 builds the Elasticsearch document id by prefixing the tenant, and it guards no null. An item with no `itemId` therefore lands on the document `<tenant>_null`. Every id-less item of that type shares one document, and each save destroys the item the previous save wrote. The call answers 200 and returns the literal string `null` as the `itemId`.

The failure is silent. Unomi logs no warning and no error, and a caller that reads the response back sees a saved item.

### Where the behaviour comes from

`ElasticSearchPersistenceServiceImpl.getDocumentIDForItemType` prefixes the tenant:

```java
private String getDocumentIDForItemType(String itemId, String itemType) {
    String tenantId = getTenantId();
    String baseId = systemItems.contains(itemType) ? (itemId + "_" + itemType.toLowerCase()) : itemId;
    return tenantId + "_" + baseId;
}
```

The same method on `unomi-2.7.x` and on `unomi-3.0.x` returns the item id unchanged, which left Elasticsearch to generate a distinct document id for each item.

`ProfileServiceImpl.save(Profile, boolean)` already refuses a null item id. The REST endpoint never reaches that guard, because it calls `saveOrMerge`, which carries no such check.

### Mechanism

Three guards now refuse the write, one per layer that can reach it:

- `ProfileServiceEndPoint.save` answers 400, because a body with no `itemId` cannot be served.
- `ProfileServiceImpl.saveOrMerge` returns `null`, the rule `save(Profile, boolean)` already applied. There is nothing to look up and nothing to merge into without an item id.
- `ElasticSearchPersistenceServiceImpl.save` and `OpenSearchPersistenceServiceImpl.save` answer `false` and log a warning, so no other caller can reach the shared document by another route.

Refusing the write is the answer rather than generating a UUID. A generated id restores the 3.0 outcome, and it hides a caller that never states the identity of what it writes.

### Effect

Measured on Apache Unomi 3.1.0-SNAPSHOT.

Before, two calls with no `itemId` both answered 200, and one document held the second item:

```
_id= default_null  {"firstName": "Second", "lastName": "Item"}
```

Now:

```
POST /cxs/profiles   (no itemId)   -> 400 {"errorMessage":"A profile states its own identity, so the body must carry an itemId"}
POST /cxs/profiles   (with itemId) -> 200
```

### Tests

`mvn test` on `persistence-spi`, `services`, `rest` and `persistence-elasticsearch/core` reports no failure.

`ProfileServiceImplTest` gains two tests: `saveOrMerge` refuses a profile with no item id, and two such profiles cannot overwrite each other.

### Notes for the reviewer

The guard sits on the save path only. `getDocumentIDForItemType` still accepts a null item id on the read path, where a null answers "not found" and destroys nothing.

This pull request is independent of https://github.com/apache/unomi/pull/864, and the two branches touch no common file.


sergehuber (changes requested):
Initial review — holding this one.

Refusing a missing `itemId` is the right 3.1 fix: persistence now prefixes `<tenant>_`, so a null `itemId` becomes `default_null` and later saves overwrite each other. Rejecting the write is better than minting a UUID (that would change the old "server assigns an id" behaviour in a different way).

Two things before merge:

1. ES and OS integration tests fail on this branch. `ProfileServiceIT.testProfileWithoutItemId` still saves a `Profile` with no `itemId`, then calls `.get(0)` on an empty search result (`IndexOutOfBoundsException` at `ProfileServiceIT.java:95`). That test needs to expect the new refusal and assert that nothing was persisted.

2. REST 400 is only on `POST /cxs/profiles`. Other item types rely on persistence `save` returning `false`. Worth deciding whether Session / Persona / similar REST create paths should also return 400, or whether a persistence-level refusal is enough.

Unit tests for the persistence guard look good.

sergehuber:
Thank you for catching this, Jonathan — this is a real 3.1 footgun and I am glad you tracked it down. Refusing a missing `itemId` is the right call: persistence now prefixes `<tenant>_`, so a null id becomes `default_null` and later saves silently overwrite each other. Rejecting the write is much better than minting a UUID, which would change the old “server assigns an id” behaviour in a different way. The persistence unit tests you added are also very welcome.

The only merge blocker I can see is that the existing integration test still describes the old behaviour. ES and OS ITs fail on `ProfileServiceIT.testProfileWithoutItemId`: it saves a `Profile` with no `itemId`, then does `.get(0)` on an empty search result (`IndexOutOfBoundsException` at line 95). A small update to that test should be enough — for example, after `saveOrMerge`, assert that no profile with that name is found (or that `save` / `saveOrMerge` returns `null` / does not persist), instead of waiting for a row and reading `.get(0)`.

A non-blocking question, only if you want to keep thinking about it in this PR: REST 400 is currently on `POST /cxs/profiles` only, and other item types rely on persistence `save` returning `false`. If you would rather keep this PR focused on the persistence guard plus the profile endpoint, that is completely fine; Session / Persona 400s can be a follow-up.

Thank you again for the careful diagnosis. Happy to re-review quickly once the IT is updated — and thank you for contributing this.

PR #866, opened 2026-09-16
UNOMI-984: Record how 3.1 answers a search condition with no resolvable type

jsinovassin:
Fixes UNOMI-984, at https://issues.apache.org/jira/browse/UNOMI-984

### What this changes

One section is added to `manual/src/main/asciidoc/migrations/migrate-3.0-to-3.1.adoc`, under "Updating applications consuming Unomi". It records how 3.1 answers a search condition whose type it cannot resolve.

### Evidence

`ProfileServiceImpl.doSearch` resolves the condition of a `Query` by its type. On 3.1 a condition with no resolvable type answers an empty list:

```java
if (query.getCondition().getConditionType() == null) {
    LOGGER.warn("Cannot execute query: condition type '{}' could not be resolved", query.getCondition().getConditionTypeId());
    return new PartialList<>();
}
```

On `unomi-3.0.x` the same body ran the query with no condition at all, which returned every item:

```java
if (query.getCondition() != null && definitionsService.resolveConditionType(query.getCondition())) {
    // run the query with the condition
} else {
    // run the query with no condition, which returns every item
}
```

Measured on 3.1.0-SNAPSHOT, against `/cxs/profiles/search/sessions`:

```
condition: {}                                 -> totalSize 0
condition: {"type":"matchAllCondition", ...}  -> totalSize 196
```

Every endpoint that takes a `Query` is affected, among them `/cxs/profiles/search`, `/cxs/profiles/search/sessions` and `/cxs/profiles/personas/search`.

### Why a note and not a revert

The two answers are opposite, and neither the request nor the response says which one the caller got. Only the server log does. Running a query whose condition could not be resolved widens the search in silence. The 3.1 answer is therefore the safer of the two, and it deserves a migration note rather than a change of behaviour.

### Mechanism

The section gives a two-column table of the 3.0 and 3.1 answers, and it names the endpoints the change reaches. The section then shows the `matchAllCondition` body a client sends to search over everything. Its place in the guide is next to the client-facing hardening section, which already records the other behaviour changes of 3.1.

### Notes for the reviewer

This pull request changes documentation only, and no code.

A client that omits the `condition` field entirely still searches over everything, so the note tells that client it needs no change.


sergehuber:
Thank you for documenting this, Jonathan. Recording the 3.1 behaviour (an unresolvable search condition now matches nothing, where 3.0 effectively matched everything) is the responsible choice, and I agree it is better to document the fail-closed answer than to restore the old “match all”.

One small coordination note: #861 also edits `migrate-3.0-to-3.1.adoc`, so it may be easiest to land that one first and then rebase this section on top. Happy to help with that if useful.

Thank you again for writing this down for operators. Docs fixes like this are very welcome.

PR #871, opened 2026-09-17
Combine the profile scope with client event filters instead of replacing it

sergehuber:
## Summary
- When listing a profile's events, always AND the profileId scope with any client-supplied filter instead of letting the filter replace the scope.
- Add an integration test that asks profile A for events while filtering on profile B and expects an empty result.

## Test plan
- [ ] `GraphQLProfileEventsScopeIT`
- [ ] CI


sergehuber:
Rebased onto #853 so the `AllITs` suite entry composes. Merge #853 first.

PR #872, opened 2026-09-17
Bound regular-expression evaluation in the matchesRegex operator

sergehuber:
## Summary
- Route `matchesRegex` through a guarded evaluation that caps pattern and input lengths, caches compiled patterns in a bounded LRU, and aborts once a fixed character-access budget is exceeded.
- Limits are configurable via `org.apache.unomi.conditions.regex.*`. Well-behaved patterns keep the same semantics.

## Test plan
- [ ] `PropertyConditionEvaluatorTest`
- [ ] `ConditionRegexSafetyIT`
- [ ] CI


sergehuber:
Rebased onto #871 so the `AllITs` suite entry composes. Merge #871 first.

PR #873, opened 2026-09-17
Simplify identifier validation patterns to a single character class

sergehuber:
## Summary
- Replace identifier patterns `^(\w|[-_@\.]){0,60}$` with the equivalent `^[\w@.-]{0,60}$` in the request, event and item schemas.
- Both forms accept the same identifiers; the single character class evaluates in linear time.
- Add `SchemaPatternSafetyTest` so shipped id patterns cannot regress to the alternation shape.

## Test plan
- [ ] `SchemaPatternSafetyTest`
- [ ] `SchemaIdPatternSafetyIT`
- [ ] CI


sergehuber:
Rebased onto #872 so the `AllITs` suite entry composes. Merge #872 first.

PR #876, opened 2026-09-17
Write filtered personalization contents back onto the request

sergehuber:
## Summary
- After sanitizing personalization contents, write the filtered list back onto the request instead of keeping the original variants.
- Keep the personalization fallback when mixed contents include a value that does not pass the filter.

## Test plan
- [ ] `ContextJsonEndpointTest`
- [ ] `ContextServletIT` mixed-contents case
- [ ] CI


sergehuber:
Rebased onto #854 so `ContextServletIT` composes. Merge #854 first.

PR #877, opened 2026-09-17
Deny public MVEL eval APIs from the scripting class loader

sergehuber:
## Summary
- Keep MVEL enabled (same as 3.0). Existing `script::` expressions still compile.
- Always deny the public MVEL eval APIs from the scripting class loader, including inner-class, array, slash, and zero-width name forms.
- Tighten forbid-pattern matching after invisible-character stripping. Request sanitizing is not treated as the control that must enumerate every prefix encoding.

## Test plan
- [ ] `SecureFilteringClassLoaderTest`
- [ ] `ExpressionFilterTest`
- [ ] `MvelScriptExecutorTest`
- [ ] CI


sergehuber:
Rebased onto #876 so later CORS / session PRs can sit on a single parent. Merge #876 first.

PR #878, opened 2026-09-18
Drop credentialed CORS on the public tracker endpoints

sergehuber:
## Summary
- Public tracker endpoints (`/cxs/context.json`, `/cxs/eventcollector`, `/cxs/client`) no longer advertise credentialed CORS.
- Add unit coverage for the public tracker CORS headers.

## Test plan
- [ ] `PublicTrackerCorsTest`
- [ ] CI

PR #880, opened 2026-09-18
Validate eventcollector query session ids and honor client IPs only from configured proxies

sergehuber:
## Summary
- Validate `sessionId` on the eventcollector query string with the same identifier rules as other public request ids.
- Honor `X-Forwarded-For` / client IP only when the request comes from a configured trusted proxy.

## Test plan
- [ ] `RequestIdentifierValidatorTest`
- [ ] `SetRemoteHostInfoActionTest`
- [ ] CI


sergehuber:
Rebased onto #878 so `EventsCollectorEndpoint` composes. Merge #878 first.

PR #881, opened 2026-09-18
Challenge on every WebSocket 401, compare the Basic scheme case-insensitively, restore idle timeout

sergehuber:
## Summary
- A WebSocket upgrade whose credential is refused now receives the same `WWW-Authenticate` challenge as a request with no credential.
- Compare the Basic scheme token case-insensitively.
- After `connection_init` authentication, restore the session's configured idle timeout instead of leaving it disabled.

Follow-up to #843 / #858 on master. The same three points are already on #844 for 3.0.x.

## Test plan
- [ ] `GraphQLServletSecurityValidatorTest`
- [ ] `SubscriptionWebSocketTest`
- [ ] CI


sergehuber:
Rebased onto #875 so the WebSocket 401 / idle-timeout follow-ups apply on top of the selected-operation and role/tenant checks, without reverting them. Merge #875 first.

